PR #82, opened 2022-03-17
Add Motoko counter example

### LINKED ISSUE #52: counter Motoko example
Recreate the counter Motoko example found here: https://github.com/dfinity/examples/tree/master/motoko/counter

Use the calc Motoko example as a template: https://github.com/demergent-labs/azle/tree/main/examples/motoko_examples/calc

A big reason for recreating the Motoko examples is to find limitations with Azle. Please document any limitations/bugs/errors you find in Azle in this issue thread.

dansteren:
This implements the Motoko counter example found here: https://github.com/dfinity/examples/tree/master/motoko/counter.

> Note: Because void types aren't yet implemented I had the update calls return the value of `counter`. This should be changed after #8 is implemented.

Resolves #52

### DIFF
--- a/LICENSE_EXTENSION.md
+++ b/LICENSE_EXTENSION.md
@@ -34,3 +34,4 @@ Your electronic signature will be considered final once the commit is included i
 | Legal Name | GitHub Username | Country of Residence | Date |
 | ---------- | --------------- | -------------------- | ---- |
 | Jordan Last | lastmjs | United States | March 14, 2022 |
+| Dan Steren | dansteren | United States | March 14, 2022 |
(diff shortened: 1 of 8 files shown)


lastmjs (commented):
This looks awesome! Void types will be coming for sure. I just want to check one thing before I approve it, did you test it locally and does it work as intended?

dansteren:
Yes. I ran `dfx deploy` and ran through the  `dfx canister call` examples in the [motoko readme](https://github.com/dfinity/examples/blob/master/motoko/counter/README.md). I also stopped dfx and restarted it to make sure that the state persisted, which it did.

The only difference was the return values because of the lack of a `void` type.

Issue #49, opened 2022-03-11
Prettier


lastmjs:

lastmjs:
@krpeacock if you want to add the prettier config let me know, ready for contributions now. Please create the most basic setup you can for now, and if you aren't interested in doing it please just let me know. Thanks!

Please note the new contributing section: https://github.com/demergent-labs/azle#contributing

PR #133, opened 2022-03-25
Linter

Raduc4:
[Issue](https://github.com/demergent-labs/azle/issues/49)
Note: uncomment `npx lint-staged` from `.husky/pre-commit` to run prettier and linter before every commit. Now it's disabled.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+# npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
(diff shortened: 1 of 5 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
lastmjs, line 4: Is this supposed to commented out per my request?

package.json
```diff
@@ -26,5 +27,16 @@
     "esbuild": "0.14.25",
     "ts-node": "10.3.1",
     "typescript": "4.4.4"
+  },
+  "devDependencies": {
+    "eslint": "^8.11.0",
```
lastmjs, line 32: If you could please remove the carrots `^`.

lastmjs (changes requested):

.husky/pre-commit
Raduc4: Yes, as we discussed, I have created a new PR, if that wasn't commented it would run prettier on my local machine and change files. Now, you can play with configurations and test when uncommenting.

lastmjs: It doesn't look commented out

lastmjs (changes requested):

.husky/pre-commit
Raduc4: Sorry, my bad

Issue #66, opened 2022-03-14
simple-to-do Motoko example


lastmjs:
Recreate the simple-to-do Motoko example found here: https://github.com/dfinity/examples/tree/master/motoko/simple-to-do

Use the calc Motoko example as a template: https://github.com/demergent-labs/azle/tree/main/examples/motoko_examples/calc

A big reason for recreating the Motoko examples is to find limitations with Azle. Please document any limitations/bugs/errors you find in Azle in this issue thread.

lastmjs:
Awesome, please review https://github.com/demergent-labs/azle#contributing and especially the https://github.com/demergent-labs/azle/blob/main/LICENSE_EXTENSION.md

tarek-eg:
Should I create a branch or fork the repo?

lastmjs:
Please create your own fork and create a feature branch off of that fork

lastmjs:
I'm going to keep the issue open so that we remember to address the remaining issues once Azle is updated

lastmjs:
There is a problem with the \n character that needs to be looked into

lastmjs:
@tarek-eg considering your work here done and merged, thanks

PR #147, opened 2022-03-31
Examples/simple to do

tarek-eg:
closes https://github.com/demergent-labs/azle/issues/66

I didn't face any issues while developing.
New line character `\n` doesn't seem to work and is treated as normal string tho.

I followed exactly the motoko example so it can be a 1 to 1 comparison.


tarek-eg:
@lastmjs ready for review

lastmjs:
I will boot this up locally and check it out, if everything works I'll merge it thanks!

tarek-eg:
One improvement we can do in the future is to add a testing setup that allow contributors to write test cases for their examples/code and run it on CI/CD by github actions, so that you don't need to always run amd verify the PR yourself.

lastmjs:
Yep I'll be putting testing infrastructure in place soon

PR #160, opened 2022-04-02
Motoko examples phone book

tarek-eg:
closes https://github.com/demergent-labs/azle/issues/63


examples/motoko_examples/phone-book/README.md
```diff
@@ -1,3 +1,69 @@
 # Phone book
 
-Reimplementation of the [Motoko Phone book example](https://github.com/dfinity/examples/tree/master/motoko/phone-book)
\ No newline at end of file
+This example demonstrates a phone book application that is accessible from your
```
lastmjs, line 3: Please just link to the original README

examples/motoko_examples/phone-book/dfx.json
```diff
@@ -0,0 +1,26 @@
...
+    "phone_book": {
+      "type": "custom",
+      "build": "npx azle phone_book",
+      "root": "src",
+      "ts": "src/phone_book/phone_book.ts",
+      "candid": "src/phone_book/phone_book.did",
+      "wasm": "target/wasm32-unknown-unknown/release/phone_book.wasm"
+    }
+  },
+  "www": {
```
lastmjs, line 12: This www object should be within the canisters object

examples/motoko_examples/phone-book/README.md
```diff
@@ -1,3 +1,69 @@
...
+1. Reserve an identifier for your canister.
+
+   ```text
+   dfx canister create --all
+   ```
+
+1. Build your front-end.
+
+   ```text
+   npm install
```
lastmjs, line 48: Same issues as with the superhoes example probably, try to make everything an exact clone as far as possible, use the asset canister

lastmjs (changes requested):
The problems here are similar to superhoes, probably stemming from the www object in the dfx.json being in the wrong place. If you can fix that and test everything, making sure it's an exact clone as far as possible, and replacing any language in the frontend app that refers to Motoko, then this should be awesome!

tarek-eg:
I think I didn't detect the www location issue because I used the npm start command instead which doesn't require the asset canister to be deployed.

.eslintrc
```diff
@@ -1,3 +1,10 @@
 {
-  "extends": ["prettier"]
-}
\ No newline at end of file
+  "extends": ["prettier"],
```
lastmjs, line 2: Let's remove this file

lastmjs (changes requested):

lastmjs:
Looking really good, if you've tested it locally from the asset canister and it works, just remove .eslintrc and I can merge this

tarek-eg:
@lastmjs this is tested locally and works fine

lastmjs:
.eslintrc file is still in there

tarek-eg:
> .eslintrc file is still in there

It was originally there, that's the one in the root

Issue #157, opened 2022-04-01
Basic dao Motoko example


tarek-eg:
There's a very useful [basic dao example ](https://github.com/dfinity/examples/tree/master/motoko/basic_dao)
Do you see any blockers on implementing that? Only thing I see is the heartbeat functionality but we can differ that to later when it's available.
It's more involved and would yield more insights on the limitations and bugs, I can start working on it if no objections.

lastmjs:
Go ahead, heartbeat won't work like you said, and there is no access to stable storage yet either. It will be interesting to see how far you get.

lastmjs:
BTW the testing infrastructure for examples should be coming today, working on it now

tarek-eg:
> BTW the testing infrastructure for examples should be coming today, working on it now

That's a good news 🥳

PR #161, opened 2022-04-02
Motoko examples Basic Dao

tarek-eg:
closes https://github.com/demergent-labs/azle/issues/157
Open questions
@lastmjs 
- [x] Is there `init` function?
- [x] How to get the `installer`? I assume if we have init function then we can get the installer = caller 
- [x] What's the difference between ic.trap and throwing an error, should we trap or throw?
- [x] Why do we have `Query` `Update` `UpdateAsync` shouldn't all query and update to be always async?


lastmjs:
No init function yet, I'll probably add init, post upgrade, pre upgrade, heartbeat all together, shouldn't be too difficult.

In init you can do ic.caller, but I am not familiar with exactly how Motoko's initialization code works under-the-hood

lastmjs:
For now you could just make a function called initialize or something that has to be called manually after the first deploy to kick things off

lastmjs:
As for trapping or throwing an error, you should notice that the Motoko DAO example never throws an error, it just returns the err variant of the Result type. You want to do the same in Azle. You'll need to create a Result type that mimics the Motoko result type: https://github.com/dfinity/examples/blob/master/motoko/basic_dao/src/Main.mo#L207

tarek-eg:
Converting logic from motoko to ts is done, I will need to test it out to check if everything works correctly, there could be a lot of issues tho.
Currently I am getting compile errors `Could not generate Candid type name for TypeScript type reference node` for this type 
```ts
export type TransferArgs = { to: Principal; amount: Tokens };
```
 I am not sure why, probably could be the same as the one I faced before, I could try to workaround it but I am sure all similar types will have the same issue, can we get this fixed?

tarek-eg:
Also I am having hard time working with generators, they are not straight forward as promises but let's see if my understanding and impl are correct.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,321 @@
...
+export function* transfer(
+  transfer: Types.TransferArgs
+): UpdateAsync<Result<void, string>> {
+  let caller = ic.caller();
+  let from_tokens = account_get(caller);
+  switch (from_tokens) {
+    case null: {
+      return yield { err: 'Caller needs an account to transfer funds' };
+    }
+    case from_tokens as Types.Tokens: {
```
tarek-eg, line 71: switch doesn't play nicely with Opt, since we need casting here, is there a better way?

lastmjs:
> Converting logic from motoko to ts is done, I will need to test it out to check if everything works correctly, there could be a lot of issues tho. Currently I am getting compile errors `Could not generate Candid type name for TypeScript type reference node` for this type and this is blocking me.
> 
> ```ts
> export type TransferArgs = { to: Principal; amount: Tokens };
> ```
> 
> I assume it could be the same issue as the one I faced before, I could try to workaround it by inlining the types but I am sure all similar types will have the same issue, can we get this fixed?

Change `Principal` to string, I don't think Principal works on `v0.0.14`. The beta has many changes and improvements, so this should work then.

lastmjs:
> Converting logic from motoko to ts is done, I will need to test it out to check if everything works correctly, there could be a lot of issues tho. Currently I am getting compile errors `Could not generate Candid type name for TypeScript type reference node` for this type and this is blocking me.
> 
> ```ts
> export type TransferArgs = { to: Principal; amount: Tokens };
> ```
> 
> I assume it could be the same issue as the one I faced before, I could try to workaround it by inlining the types but I am sure all similar types will have the same issue, can we get this fixed?

It also could be that you are using `Array`. That is not supported, use the `[]` syntax instead. To be sure, change all uses of `Prinicpal` to string and all uses of `Array<T>` to `T[]`.

lastmjs:
Also, you need to use the azle type `Variant`, not the TypeScript type `enum` for Candid variants.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,334 @@
...
+      account_put(caller, { amount_e8s: from_amount });
+      return { ok: void 0 };
+    }
+  }
+
+  return { err: 'Unexpected error' };
+}
+
+/// Execute all accepted proposals
+function* execute_accepted_proposals(): Generator<
```
lastmjs, line 279: This function will be the only one that needs to return a generator, along with execute_proposal. But cross canister calls are not coming until beta (which is extremely imminent).

Async should become much clearer with the beta and the new examples that I'll release. There will also be a lot of documentation coming.

lastmjs (changes requested):

lastmjs:
> Why do we have Query Update UpdateAsync shouldn't all query and update to be always async?

In the underlying JavaScript engine, there is no need for asynchronous code unless you are performing an asynchronous operation. So for `Query` and `Update` functions that do not call any of the asynchronous IC APIs, there is no need for them to be asynchronous. In fact if they're made asynchronous it might cause issues.

It's only when you do something like `ic.rawRand` or cross-canister calls, any asynchronous API that is provided to you from `azle` basically, that you would need to deal with generators, and the `UpdateAsync` and `Async` types provide azle with that information and give you the types to represent what's happening.

In the beta things have gotten pretty simple with generators, for example a call to `rawRand` looks like this:

```typescript
import { UpdateAsync, ic } from 'azle';

export function* randomness(): UpdateAsync<nat8[]> {
  return yield ic.rawRand();
}
```

You can also call `ic.rawRand` a few functions down, and I believe only the top level function has to `yield`. It'll all become clearer and simpler with the beta.

Oh and there is no `QueryAsync` because `Query` functions cannot perform cross-canister calls, which is a limitation of the IC currently.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,86 @@
...
+  proposal_submission_deposit: Tokens;
+};
+export type BasicDaoStableStorage = {
+  accounts: Account[];
+  proposals: Proposal[];
+  system_params: SystemParams;
+};
+
+export function accounts_fromArray(arr: Account[]): Map<Principal, Tokens> {
+  var s = new Map<Principal, Tokens>();
```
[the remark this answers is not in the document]
tarek-eg: Just copy pasta from motko

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,86 @@
+import { Principal, int32 } from 'azle';
+
+export type Account = { owner: Principal; tokens: Tokens };
```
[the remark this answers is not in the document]
tarek-eg: It is available in 0.0.14 but the type is actually a string, will use a string to be safe here.

tarek-eg:
That was great removing the generator stuff simplified things a lot.
I am still getting compile errors at this type 
```ts
export type Tokens = {
  amount_e8s: int32;
};
```

lastmjs:
I'm working on `init` and `heartbeat` now

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,338 @@
...
+/* stable */ let next_proposal_id: int32 = 0; // TODO should be nat
+/* stable */ let system_params: Types.SystemParams = init.system_params;
+
+// function heartbeat() {
+//   await execute_accepted_proposals();
+// };
+
+// simulate init function
+// for now as a hack we need to call this function manually
+export function initialize(init: Types.BasicDaoStableStorage) {
```
lastmjs, line 38: You might (probably will) run into uses using the types like you are using them, `Types.WhateverType`. Support for that can be added later. Right now you need to use the type directly, so just import `import { WhateverType }` instead of storing all exports in a variable.

Basically import and use each type individually.

lastmjs (changes requested):

examples/motoko_examples/basic-dao/src/basic_dao.ts
lastmjs: If it does work then that's great, but I doubt it will. Just a warning

tarek-eg: I will give it a try

tarek-eg:
I have commented all the code that causes compile errors, I will need to narrow down the actual issue.

lastmjs:
Init now works on master

tarek-eg:
Looks like there's an issue with init function when it gets passed a principal as an argument
Ex.
`dfx deploy --argument "(principal \"3su3b-zm3rt-5m634-pf5zx-23wy4-iuqgi-vt6vq-bpgs5-q2kb5-lwmfx-pae\")"`
```
Error: Invalid data: Unable to serialize Candid values: type mismatch: principal "3su3b-zm3rt-5m634-pf5zx-23wy4-iuqgi-vt6vq-bpgs5-q2kb5-lwmfx-pae" cannot be of type text
```

lastmjs:
I'll get to this first thing

tarek-eg:
> Are you using dfx 0.9.3 and azle 0.1.0? I am not finding passing in principals to be an issue with those versions.

Let me check again, passing a principal as argument works but it seems is error is originated from somewhere else

tarek-eg:
I changed the status to open for review to be able to run the tests.

lastmjs:
You also need to add the path here: https://github.com/demergent-labs/azle/blob/main/.github/workflows/test.yml#L17

And if you could put your addition in alphabetical order, `motoko_examples/basic_dao` should be the first directory of the `motoko_examples`

lastmjs:
Also you can run the tests locally from the root directory of your example `npm test`.

tarek-eg:
> Also you can run the tests locally from the root directory of your example `npm test`.

Thanks for the tip.

tarek-eg:
@lastmjs Is there a way to execute bash commands in tests?
For example I need to simulate different identities in tests so I can do:
`dfx identity new Alice; dfx identity use Alice;`
`
dfx canister call basic_dao transfer "(record {to = principal \"rrkah-fqaaa-aaaaa-aaaaq-cai\"; amount = record {amount_e8s = 1000} })"
`

lastmjs:
You can't do that right now but I can add that

tarek-eg:
> The tests can't do that right now but I can add that, that will probably solve some issues I had in other tests as well

That would be nice.
It's not a blocker but there are some useful tests for this example that can't be done without that.

tarek-eg:
@lastmjs I think there's an issue with `CanisterResult` type or specifically the generic part.

Try to deploy this example locally
```
dfx deploy basic_dao --mode=reinstall --argument "(vec { record { owner = principal \"rrkah-fqaaa-aaaaa-aaaaq-cai\"; tokens = record { amount_e8s = 100_000_000 }; }; record { owner = principal \"vsane-vq2o4-awwmq-qlf3d-253a4-yf3nz-rqz6a-4n5se-we6yw-bmdag-cae\"; tokens = record { amount_e8s = 100_000_000 }; };}, vec {},record { transfer_fee = record { amount_e8s = 10_000 };proposal_vote_threshold = record { amount_e8s = 10_000_000 }; proposal_submission_deposit = record { amount_e8s = 10_000 };})"
```

The deployment should fail.

To make it work
change the return type of `_transferTokens`
from
```ts
export function _transferTokens():  Update<CanisterResult<void>>{
  return { err: 'Not implemented' };
}
```
to 
```ts
export function _transferTokens(): Update<Res> {
  return { err: 'Not implemented' };
}
```

where `Res` is just 
```ts
type Res = Variant<{
  ok?: nat;
  err?: string;
}>;
```

lastmjs:
You can't return generics in your Query or Update calls, they have to be statically defined. This is a limitation of Candid/IC not of Azle. Generics can be used in your own TypeScript code, but when it comes to the parameters or return types of any functions exposed to the IC (anything that needs to be Candid serializable/deserializable) then you must have static types.

Here is the Azle documentation on all of the available types: https://github.com/demergent-labs/azle#candid-data-types

I will make it clear that you cannot use generics or any kind of dynamic code in the Candid TypeScript types

lastmjs:
BTW, we will probably build an Azle VS Code plugin that does more static analysis than TypeScript so that we can catch all of these issues for developers.

tarek-eg:
I think all works now, I will need to use the stable storage api.

tarek-eg:
There's an issue with the tests where results that span multiple lines are failing

```bash
Running test basic_dao getSystemParams
 test basic_dao getSystemParams failed
 Expected: (
record {
transfer_fee = record { amount_e8s = 10_000 : nat };
proposal_vote_threshold = record { amount_e8s = 10_000_000 : nat };
proposal_submission_deposit = record { amount_e8s = 10_000 : nat };
},
), Received: (
  record {
    transfer_fee = record { amount_e8s = 10_000 : nat };
    proposal_vote_threshold = record { amount_e8s = 10_000_000 : nat };
    proposal_submission_deposit = record { amount_e8s = 10_000 : nat };
  },
)
```

tarek-eg:
Other than that I think it's ready for review.
I have tested most of the interfaces manually and they look fine.
I tried upgrading the canister to test the stable storage but it didn't work for me the canister was returning an error, probably I am doing something wrong.

lastmjs:
If you look at some of the other tests, you just have to make the formatting perfect. Just use \n and the correct number of spaces. Each line is indented by two more spaces from the previous line

lastmjs:
Reviewing this now, sorry it took so long

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+  ic.print(`caller ${ic.caller()}`);
+  ic.print(`accounts ${ic.stableStorage<StableStorage>().accounts}`);
+  return accountGet(ic.caller()) ?? zeroToken;
+}
+
+// Lists all accounts
+export function listAccounts(): Query<Account[]> {
+  // this is causing errors
+  // return Array.from(accounts, ([owner, tokens]) => ({ owner, tokens }));
+  const arr: Account[] = [];
```
lastmjs, line 137: Please do this immutably using `Array.reduce` or `Map.values`.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+  return arr;
+}
+
+// / Submit a proposal
+// /
+// / A proposal contains a canister ID, method name and method args. If enough users
+// / vote "yes" on the proposal, the given method will be called with the given method
+// / args on the given canister.
+export function submitProposal(payload: ProposalPayload): Update<Res> {
+  let caller = ic.caller();
```
lastmjs, line 154: Prefer const for immutability, use a `SubmitProposalResult` type.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+/// Return the proposal with the given ID, if one exists
+export function getProposal(proposal_id: nat): Query<Opt<Proposal>> {
+  return proposalGet(proposal_id);
+}
+
+/// Return the list of all proposals
+export function listProposals(): Query<Proposal[]> {
+  // this is causing errors
+  //return Array.from(proposals, ([_, proposal]) => proposal);
+  const arr: Proposal[] = [];
```
lastmjs, line 186: Please do this immutably using `Array.reduce` or `Map.values`.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+  ic.stableStorage<StableStorage>().proposals.forEach((proposal) => {
+    arr.push(proposal);
+  });
+
+  return arr;
+}
+
+type VoteResult = Variant<{
```
lastmjs, line 194: Let's get this into the `types.ts` file.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+      proposal.voters.push(caller);
+      const voters = proposal.voters;
+
+      if (
+        votes_yes >=
+        ic.stableStorage<StableStorage>().systemParams.proposal_vote_threshold
+          .amount_e8s
+      ) {
+        // Refund the proposal deposit when the proposal is accepted
+        const account = accountGet(proposal.proposer)!;
```
lastmjs, line 240: We should handle a `null` account here, let's get rid of the ending `!`.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+
+      let votes_yes = proposal.votes_yes.amount_e8s;
+      let votes_no = proposal.votes_no.amount_e8s;
+      if ('Yes' in args.vote) {
+        votes_yes += account.amount_e8s;
+      }
+      if ('No' in args.vote) {
+        votes_no += account.amount_e8s;
+      }
+      proposal.voters.push(caller);
```
lastmjs, line 231: I think we should create a new array here and do the mutation at the end. Please check what the List.push in Motoko does. If it pushes into an array creating a new copy, then I think we should do that (as we want to manage our mutations and do them all at once if possible).

If that's not what Motoko does, then I suppose we can leave it.

I'm trying to balance best practices with just following what the Motoko code does. I think the Motoko code could be improved in many ways but I want the Azle Motoko examples to be mirrors of the Motoko examples.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+}
+
+/// Update system params
+///
+/// Only callable via proposal execution
+export function updateSystemParams(
+  payload: UpdateSystemParamsPayload
+): Update<void> {
+  let caller = ic.caller();
+  if (caller != ic.id()) {
```
lastmjs, line 289: As a best practice always use `===` and `!==` in place of `==` and `!=` in JavaScript.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+    return;
+  }
+  ic.stableStorage<StableStorage>().systemParams = {
+    ...ic.stableStorage<StableStorage>().systemParams,
```
lastmjs, line 293: This will probably need to be reworked once `UpdateSystemParamsPayload` uses optional types.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  Succeeded?: null;
+  // Enough "yes" votes have been cast to accept the proposal, and it will soon be executed
+  Accepted?: null;
+}>;
+export type Tokens = {
+  amount_e8s: nat; // TODO should be nat
+};
+// export type TransferArgs = { to: Principal; amount: Tokens };
+export type UpdateSystemParamsPayload = {
+  transfer_fee?: Tokens;
```
lastmjs, line 39: You must use `Opt` if you expect a value to be null in any Candid record type.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+      account_put(caller, { amount_e8s: fromAmount });
+      return { ok: 0n };
+    }
+  }
+  return { err: 'Unexpected error' };
+}
+
+/// Execute all accepted proposals
+function executeAcceptedProposals(): void {
+  const allProposals: Proposal[] = [];
```
lastmjs, line 324: Please be immutable here with `Array.reduce` or `Map.values`.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+    return;
+  }
+  ic.stableStorage<StableStorage>().systemParams = {
+    ...ic.stableStorage<StableStorage>().systemParams,
+    ...payload,
+  };
+}
+
+/// Deduct the proposal submission deposit from the caller's account
+function deductProposalSubmissionDeposit(caller: Principal): Res {
```
lastmjs, line 299: Or you can make a `Res` type that is just `ok? null; err?: string` as well, if you don't want `ok` to have a value.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+    }
+  }
+}
+
+// TODO This should be async
+// Execute the given proposal
+function executeProposal(proposal: Proposal): Res {
+  try {
+    // let payload = proposal.payload;
+    // TODO we need raw cross canister calls ic.callRaw()
```
lastmjs, line 352: Yep I need to get `call_raw` implemented...should be pretty easy.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+  state: ProposalState;
+  timestamp: nat;
+  proposer: Principal;
+  votes_yes: Tokens;
+  payload: ProposalPayload;
+};
+export type ProposalPayload = {
+  method: string;
+  canister_id: Principal;
+  message: string;
```
lastmjs, line 17: This should be a `nat8[]`, which is the equivalent of a Motoko `Blob`.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export type ProposalState = Variant<{
+  // A failure occurred while executing the proposal
+  Failed?: string;
```
lastmjs, line 22: If Motoko has all of these starting lowercase then I think we should do the same in TypeScript.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+export type Tokens = {
+  amount_e8s: nat; // TODO should be nat
+};
+// export type TransferArgs = { to: Principal; amount: Tokens };
+export type UpdateSystemParamsPayload = {
+  transfer_fee?: Tokens;
+  proposal_vote_threshold?: Tokens;
+  proposal_submission_deposit?: Tokens;
+};
+export type Vote = Variant<{
```
lastmjs, line 43: I would think we should lowercase these, and also variants should have a `?` as part of the type, otherwise you would have to put both properties.

examples/motoko_examples/basic-dao/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export function accounts_fromArray(arr: Account[]): Map<Principal, Tokens> {
+  const s = new Map<Principal, Tokens>();
+  for (let account of arr) {
+    s.set(account.owner, account.tokens);
+  }
+  return s;
+}
+export function proposals_fromArray(arr: Proposal[]): Map<nat, Proposal> {
+  const s = new Map<
```
lastmjs, line 76: Since `s` is intended to be mutated let's use `let`

lastmjs (changes requested):

tarek-eg:
I will work on the comments in a couple of weeks

lastmjs:
I've created a new test system that is drastically simpler to use and much more flexible, should be out by the time you get back to this

lastmjs:
v0.2.2 is out with a completely revamped testing infrastructure. You should be able to do anything you need to do now, take a look at the examples.

tarek-eg:
While everything is working almost fine, probably the most pressing issue now is the DX.
I have been debugging a failed build for some time now with no luck.
It would be nice if we have an eslint plugin or at least a better error messages to know exactly where is the problematic code.

lastmjs:
Yes a VS Code plugin combined with much better compiler errors are coming projects

examples/motoko_examples/basic-dao/src/basic_dao.ts
tarek-eg: In Motoko they are always creating a new array instead of mutating, I also prefer immutability in general.

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+export function heartbeat(): Heartbeat {
+  executeAcceptedProposals();
+}
+
+function accountGet(id: Principal): Opt<Tokens> {
+  return ic.stableStorage<StableStorage>().accounts.get(id) ?? null;
+}
+
+function account_put(id: Principal, tokens: Tokens) {
+  ic.stableStorage<StableStorage>().accounts.set(id, tokens);
```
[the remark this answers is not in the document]
tarek-eg: I am not sure I understand this correctly.
Since the Principal type is still of type string this should be fine?

lastmjs:
`dfx generate`, you just run that from the root directory of your example. Make sure you add the declarations path to your dfx.json. Also once you generate the code, you need to go to dfx_generated/canister_name/index.js and comment out the last line that has `createActor`. Look at the other examples and you should see what I mean.

tarek-eg:
@lastmjs what's the correct way to call an async function from within heartbeat?
It seems heartbeat is not executing `executeAcceptedProposals` and showing error 
`Panicked at 'called Option::unwrap() on a None value', src/src/lib.rs:1474:65`

I have tried many permutation from waiting to non waiting `yield` but nothing seems to be working.
I have also tried just a simple async `function*` but still not working.

tarek-eg:
> I'm sorry you ran into this without any good documentation. This is a major issue I discovered while implementing the Multisig Vault. What you are doing should be possible, basically you would turn the heartbeat method into a generator and yield every other generator, just like `async/await` but instead `*/yield`.
> 
> Unfortunately, the Azle's generator handling currently can't deal with multiple levels of generators. So, the answer is to get rid of the nice function abstractions that you have and do everything from within the heartbeat method itself. If you yield (like for a cross-canister call), you basically need to yield in the top-level canister method.
> 
> So you should be able to get this to work now but just inlining all off the code in the heartbeat method. I just need to update the generator handling to recursively handle yielding of generators.
> 
> See here: #203

Thanks, that did the trick.
All tests now should pass as they all succeeded locally.
You can give it another review.

examples/motoko_examples/basic-dao/test/test.js
```diff
@@ -0,0 +1 @@
+require('azle/test/test.js');
```
lastmjs, line 1: This file can now be deleted

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,374 @@
...
+// /* stable */ let installer: Opt<Principal> = null;
+// /* stable */ let accounts = accounts_fromArray([]);
+// /* stable */ let proposals = proposals_fromArray([]);
+// /* stable */ let nextProposalId: nat = 0n;
+// /* stable */ let systemParams: SystemParams = {
+//   transfer_fee: { amount_e8s: 0n },
+//   proposal_vote_threshold: { amount_e8s: 0n },
+//   proposal_submission_deposit: { amount_e8s: 0n },
+// };
+
```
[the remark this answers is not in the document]
lastmjs: It doesn't look like this was resolved. The stable types must not have dynamic keys.

See these examples for how to use stable storage:

https://github.com/demergent-labs/azle/blob/main/examples/stable_storage/src/stable_storage.ts
https://github.com/demergent-labs/azle/blob/main/examples/pre_and_post_upgrade/src/pre_and_post_upgrade.ts

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,385 @@
...
+export function accountBalance(): Query<Tokens> {
+  return accountGet(ic.caller()) ?? zeroToken;
+}
+
+// Lists all accounts
+export function listAccounts(): Query<Account[]> {
+  const accountList = Object.keys(accounts).map((owner) => ({
+    owner,
+    tokens: accounts[owner],
+  }));
```
lastmjs, line 154: Is there a reason for the extra set of parentheses around the body of the lambda?

lastmjs: Oh I see, it's shorthand.

lastmjs (changes requested):

examples/motoko_examples/basic-dao/src/basic_dao.ts
```diff
@@ -0,0 +1,385 @@
...
+  // Update proposal state, so that it won't be picked up by the next heartbeat
+  for (let proposal of acceptedProposals) {
+    updateProposalState(proposal, { executing: null });
+  }
+
+  for (let proposal of acceptedProposals) {
+    try {
+      const payload = proposal.payload;
+
+      let res = yield ic.call_raw(
```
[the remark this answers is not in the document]
tarek-eg: Would that work?
```ts
const res: CanisterResult<Variant<{ ok: null; err: null }>> =
        yield ic.call_raw(
          payload.canister_id,
          payload.method,
          payload.message,
          0n
        );
```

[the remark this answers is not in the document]
tarek-eg: Yes should be const, just slipped.

tarek-eg:
@lastmjs 
The basic integration tests step is failing, do you know what might be the reason?

Issue #213, opened 2022-04-11
Azle and @dfinity/agent TypeScript types don't quite match


lastmjs:
- [x] Principal: https://github.com/demergent-labs/azle/issues/251
- [ ] Variants
- [ ] Opt

lastmjs:
We are just going with DFINITY's way of doing things for now. We aren't really concerning ourselves with this. If anything, I would like to move away from Candid and embrace WIT (WebAssembly Interface Types).

Issue #211, opened 2022-04-11
Principal and Variant types in Azle do not match the dfinity/agent generated types


lastmjs:
I could probably use the same Principal code in Azle that is used in @dfinity/agent, but I prefer the way Azle is doing variants, it is more like what Candid does, or so I thought...

lastmjs:
Duplicate: https://github.com/demergent-labs/azle/issues/213

PR #233, opened 2022-04-18
Unify principal type

krpeacock:
Surprisingly, this works out of the box, according to the existing test suite

Relates to #211 #213

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "homepage": "https://github.com/lastmjs/azle#readme",
   "dependencies": {
+    "@dfinity/principal": "^0.11.0",
     "@swc/core": "1.2.151",
     "azle-syn": "0.0.0",
     "esbuild": "0.14.25",
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,5 @@
 {
 	"singleQuote": true,
-	"quoteProps": "consistent"
+	"quoteProps": "consistent",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
   },
   "homepage": "https://github.com/lastmjs/azle#readme",
   "dependencies": {
+    "@dfinity/principal": "^0.11.0",
```
lastmjs, line 26: Please remove the `^` here

lastmjs (changes requested):
Did you run the test suite locally?

lastmjs:
Please follow the steps in the contributors section here: https://github.com/demergent-labs/azle#contributing

lastmjs:
I'll be very surprised if these tests pass, because Boa doesn't have class support yet. If it doesn't pass I think I'd like to close this issue and implement this later on when I add the version of boa with class support.

I am really trying to keep contributions contained to good first issues and help wanted for now.

lastmjs:
So I think I know what's going on here, `import type { Principal } from '@dfinity/principal';` is just importing a type declaration, which is compiled away. None of the runtime code has been affected. Does that sound correct to you? I'm pretty sure this is more involved, and I'd preferred Demergent Labs to just take it over from here.

lastmjs:
If you're interested in working on this in a bit once Azle is ready let me know, I can prepare an issue with the necessary information once the time comes.

PR #260, opened 2022-05-11
Add motoko quicksort example

dansteren:
This implements the Motoko quicksort example found here: https://github.com/dfinity/examples/tree/master/motoko/quicksort.

I wrote a test for it as well using the values from the example in the motoko example's [readme](https://github.com/dfinity/examples/blob/master/motoko/quicksort/README.md) and the test passes.

Resolves https://github.com/demergent-labs/azle/issues/65


examples/motoko_examples/quicksort/src/motoko_base.ts
```diff
@@ -0,0 +1,29 @@
+import {
+    int
+} from 'azle';
+
+export const Int = {
+    abs: Math.abs,
```
lastmjs, line 6: Let's also pull these out as named functions with the `function` keyword syntax

lastmjs (changes requested):

examples/motoko_examples/quicksort/src/motoko_base.ts
dansteren: I've pulled out the `compare` function and didn't see any other offenses in the rest of the codebase. This should be good now.

lastmjs:
Everything is now up-to-date with the github actions flow, so if you could add the path to your tests in the following places (make sure to maintain the alphabetical order):

- [ ] azle/.github/workflows/test.yml
- [ ] test/test.sh
- [ ] publish.sh

lastmjs (approved):

Issue #271, opened 2022-05-13
Expose installer information on the global ic object


dansteren:
### User Story

**An a** developer writing a TS/JS app with Azle,
**I want** a way to access the ID of the principal that installed the canister from within my canister functions
**So that** I can expose it through my API or base logic off of it.

---

When implementing #68 I discovered that we weren't exposing this information. That example needs access to the principal that installed the canister but there is likely other information stored as well. We should expose it all.

This should likely be exposed on the global `ic` object as `ic.installer`.

dansteren:
This user story can already be accomplished by manually storing the `ic.caller()` in a variable from within the Init function like so:

```ts
let installer: Principal;

export function init(): Init {
  installer = ic.caller();
}
```
As I understand it, Motoko merely gives you some syntactic sugar for doing this by way of binding the argument to the `this` keyword. Alternatively you could manually store the variable like we do in JS, both are functionally equivalent. The former is just a shorthand. To recap, I believe the following to examples are equivalent:

1. Binding the message to `this` and making it accessible to all functions in the future:
```motoko
shared (message) actor class MyCanister = this {
  public query func installer() : async Principal {
    return message.caller;
  };
}
```

2. Or explicitly storing the message from the initialization in a variable for access later:
```motoko
shared (message) actor class MyCanister {
  let owner = message.caller;
  public query func installer() : async Principal {
    return owner.caller;
  };
}
```

TLDR; we shouldn't implement this.

See https://smartcontracts.org/docs/current/developer-docs/build/languages/motoko/caller-id/
See https://github.com/demergent-labs/azle/pull/275#issuecomment-1126650319

dansteren:
To completely backtrack on my last comment, I believe that a case can still be made for exposing the canister installer on the global `ic` object.

After my last comment I continued working on #68. While running further tests I found that there are some nuances in motoko that surface on re-deploys.

Consider the following motoko canister:
```motoko
import Principal "mo:base/Principal";

shared (msg) actor class MyCanister = this {
  public query func installer() : async Principal {
    return msg.caller;
  };
};
```

After running `dfx deploy` a single time, calling `dfx canister call my_canister installer` would return the principal that installed the canister.

Now suppose you:
1. create a new identity
2. make it a controller of the canister
3. add another method (so that the hash changes and you can redeploy)
4. redeploy the canister from the new identity
5. run `dfx canister call my_canister installer`

[...]

PR #275, opened 2022-05-14
Add motoko whoami example

dansteren:
This implements the Motoko whoami example found here: https://github.com/dfinity/examples/tree/master/motoko/whoami.

I have tests for each method that the canister exposes.

The `installer` method cannot be implemented until #271 is resolved.

One question I had was whether I should make the `idQuick` method a `Query` or an `Update`. The [motoko repo](https://github.com/dfinity/examples/blob/master/motoko/whoami/src/Main.mo#L30) does not mark it as a `Query` but I think that was just an oversight on their part. Both codebases work as a query call. So, I'm not sure if I should assume that's what they meant, or strictly match them.

Resolves https://github.com/demergent-labs/azle/issues/68


tarek-eg:
> The `installer` method cannot be implemented until #271 is resolved.

Can't you just use the `init` lifecycle hook?
```js
export function init(): Init {
  installer = ic.caller();
}
```

I think this is the correct approach, you can find a full example in my open PR.

examples/motoko_examples/whoami/test/test.ts
```diff
@@ -0,0 +1,104 @@
...
+    run_tests,
+    Test
+} from 'azle/test';
+import { execSync } from 'child_process';
+import { createActor } from './dfx_generated/whoami';
+import { Ed25519KeyIdentity } from '@dfinity/identity';
+import { SignIdentity } from '@dfinity/agent';
+
+function createIdentity(seed: number): SignIdentity {
+    const seed1 = new Array(32).fill(0);
```
lastmjs, line 11: Please explain this to me

lastmjs:
@tarek-eg We're wondering if `installer` is some sort of built-in Motoko property that we should emulate in Azle as well.

tarek-eg:
> @tarek-eg We're wondering if `installer` is some sort of built-in Motoko property that we should emulate in Azle as well.

It's not something built-in in motoko, but they provide a way to get the principal that installed the canister `owner`.
```ts
shared(msg) actor class Counter(init : Nat) {
  // ... msg.caller ... === installer (owner)
}
```
in shared functions

```ts
shared(msg) func inc() : async () {
  // ... msg.caller ...
}
```

in rust you don't have the same but you can simulate that by storing the caller in init function which is effectively the installer.

examples/motoko_examples/whoami/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "scripts": {
+        "test": "ts-node --transpile-only --ignore=false test/test.ts"
+    },
+    "dependencies": {
+        "@dfinity/identity": "^0.11.1",
```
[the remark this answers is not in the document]
dansteren: Done

examples/motoko_examples/whoami/src/whoami.ts
```diff
@@ -0,0 +1,68 @@
...
+    id(): CanisterResult<Principal>;
+    idQuick(): CanisterResult<Principal>;
+}>;
+
+type WhoAmIResult = Variant<{
+    ok: Principal;
+    err: string;
+}>;
+
+type StableStorage = Stable<{
```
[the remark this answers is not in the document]
dansteren: When testing it appeared that the storage remained stable. However, I think this was an artifact of me not changing the code before re-deploying. I will check the motoko example again and make sure that is the case and then likely remove this stable storage call. Thanks for pointing this out :)

examples/motoko_examples/whoami/src/whoami.ts
```diff
@@ -0,0 +1,68 @@
...
+    return ic.caller();
+}
+
+// Return the principal identifier of this canister.
+export function* id(): UpdateAsync<Principal> {
+    const thisCanister = ic.canisters.WhoAmICanister<WhoAmICanister>(ic.id());
+
+    const result: WhoAmIResult = yield thisCanister.whoami();
+
+    return result.ok || ""
```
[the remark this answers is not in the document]
dansteren: Done

examples/motoko_examples/whoami/test/test.ts
dansteren: We talked about this offline. I was just following the precedence set by @tarek-eg. We explored a different way to do this without mutating so I will make some slight changes, but the rough idea is to just generate a random seed for the identity function.

dansteren:
@tarek-eg thanks for pointing this out and breaking it down. I misunderstood when I was reading the motoko example and thought it was built into the language. Thanks for helping me to understand otherwise. I'll go back through and store the installer in the init message.

dansteren:
@lastmjs these example have the notices/readmes already saved into directories in the codebase. So they're not showing up in the PR because they already exist. But I did double check and they're both there so we should be good.

examples/motoko_examples/whoami/src/whoami.ts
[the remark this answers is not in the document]
dansteren: I re-tested this making sure to actually modify the motoko code in between deploys. As expected, it did not persist. I've updated the PR and removed the call to stable storage.

dansteren:
@lastmjs this is ready for review again.

examples/motoko_examples/whoami/src/whoami.ts
```diff
@@ -0,0 +1,73 @@
...
+    id(): CanisterResult<Principal>;
+    idQuick(): CanisterResult<Principal>;
+}>;
+
+type WhoAmIResult = Variant<{
+    ok: Principal;
+    err: string;
+}>;
+
+let install: Principal;
```
lastmjs, line 26: These types aren't quite correct if you think about it. I think we should type these as `Opt<Principal>` and set their initial values to null. Right now their initial values are `undefined` so the type should actually be `Principal | undefined` but it isn't.

examples/motoko_examples/whoami/src/whoami.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// Manually re-save these variables after new deploys.
+export function postUpgrade(somebody: Principal): PostUpgrade {
+    install = ic.caller();
+    someone = somebody;
+}
+
+// Return the principal identifier of the wallet canister that installed this
+// canister.
+export function installer(): Query<Principal> {
```
lastmjs, line 43: We might want these functions to return `Opt<Principal>`. The typing is slightly different than Motoko because of the Azle environment, we don't know that init and postUpgrade will set these values so we kind of need the `Opt` type.

examples/motoko_examples/whoami/test/test.ts
```diff
@@ -0,0 +1,135 @@
...
+    },
+    {
+        // TODO hopefully we can get rid of this: https://forum.dfinity.org/t/generated-declarations-in-node-js-environment-break/12686/16?u=lastmjs
+        name: 'waiting for createActor fetchRootKey',
+        wait: 5000
+    },
+    {
+        name: 'deploy',
+        prep: async () => {
+            execSync(`dfx deploy --argument '(principal "${installationPrincipal}")'`, {
```
lastmjs, line 48: I feel like it would be less confusing if we had an `installationPrincipal` and a `someonePrincipal` to match what the Motoko example is doing. As we've talked about it I felt like it was confusing to follow when we were talking about the installer and when we were just talking about the argument being passed in (someone in the Motoko example).

examples/motoko_examples/whoami/test/test.ts
```diff
@@ -0,0 +1,135 @@
...
+        test: async () => {
+            const result = await whoami_canister.id();
+
+            return {
+                ok: result.toString() === canisterId
+            };
+        }
+    },
+    {
+        name: 'idQuick',
```
lastmjs, line 94: If you haven't already if you could create an issue telling is to consider creating the equivalent of `Principal.fromActor`.

lastmjs (changes requested):

examples/motoko_examples/whoami/test/test.ts
dansteren: Created https://github.com/demergent-labs/azle/issues/333

examples/motoko_examples/whoami/test/test.ts
dansteren: I've created a new `someonePrincipal` so that there isn't confusion between these two.

examples/motoko_examples/whoami/src/whoami.ts
```diff
@@ -0,0 +1,75 @@
...
+    idQuick(): CanisterResult<Principal>;
+}>;
+
+type WhoAmIResult = Variant<{
+    ok: Principal;
+    err: string;
+}>;
+
+// Initialize the variables to ensure that they aren't `undefined`.
+// We use the zero address but any principal could be used.
```
lastmjs, line 27: This comment may not technically correct, it's the zero principal not the zero address. Addresses are different concepts and this could lead to confusion.

lastmjs (changes requested):
Everything is looking really good, if we could just make that principal/address comment more correct

examples/motoko_examples/whoami/src/whoami.ts
dansteren: Updated `address` to `principal`. We should be good now.

lastmjs (approved):

PR #290, opened 2022-05-17
Add hello-world motoko example

### LINKED ISSUE #59: hello-world Motoko example
Recreate the counter Motoko example found here: https://github.com/dfinity/examples/tree/master/motoko/hello-world

Use the calc Motoko example as a template: https://github.com/demergent-labs/azle/tree/main/examples/motoko_examples/calc

A big reason for recreating the Motoko examples is to find limitations with Azle. Please document any limitations/bugs/errors you find in Azle in this issue thread.

dansteren:
This implements the Motoko hello-world example found here: https://github.com/dfinity/examples/tree/master/motoko/hello-world.

Note: the only thing this canister does is print output into the terminal where the ic replica is running. I'm not sure at the moment how to monitor that output from within a test so I'm just leaving out tests completely.

Resolves #59


lastmjs (changes requested):
Tests will still be useful so that we know the canister compiles, deploys, and runs. Here is an example of a test for print: https://github.com/demergent-labs/azle/blob/main/examples/ic_api/test/test.ts#L71

It just calls the function, but I think the test is still useful because we at least know the function can be invoked and calling print does not error out.

tarek-eg:
> Tests will still be useful so that we know the canister compiles, deploys, and runs. Here is an example of a test for print: https://github.com/demergent-labs/azle/blob/main/examples/ic_api/test/test.ts#L71
> 
> It just calls the function, but I think the test is still useful because we at least know the function can be invoked and calling print does not error out.

Regarding tests, I was wondering if we should run the tests always against the current azle code in every PR instead of the published one.

The reason is that if some one introduced a bug/ breaking change to azle in some PR the tests in the examples will still pass since they are pinned to a specific version and fetched from npm.

Or maybe you've already taken care of that in the CI/CD pipeline?

lastmjs:
@tarek-eg That's exactly what we're doing :) Feature branches run against the azle code in the repository, while release branches will run against the repository and the npm version.

tarek-eg:
> @tarek-eg That's exactly what we're doing :) Feature branches run against the azle code in the repository, while release branches will run against the repository and the npm version.

I suspected that was the case but just was wondering as I know we are pinning azle version in every example.

tarek-eg:
> @tarek-eg That's exactly what we're doing :) Feature branches run against the azle code in the repository, while release branches will run against the repository and the npm version.

I suspected that was the case but just was wondering as I know we are pinning azle version in every example.

lastmjs (approved):

lastmjs (dismissed):

lastmjs (approved):

PR #367, opened 2022-05-31
Ds/60 life

dansteren:
This is implemented up through v1. Basically we hit some bugs with migrations. Migrating from a canister without stable storage, to adding stable storage in v1 broke it. We couldn't deploy. So, we need to fix `Migrate` and then come back to this, make the transition from v0 to v1 nice, and then work on v2.

Also, you should check the declarations from v1 into source control so that they're there when you move it over to `src`.


dansteren:
We need to go back and finish this out. Closing for now.

PR #370, opened 2022-06-09
Null

### LINKED ISSUE #200: Implement null type
I do not think we have full support for the null type, just Opt: https://smartcontracts.org/docs/candid-guide/candid-types.html#type-null

- [x] Candid
- [x] AzleIntoJsValue
- [x] AzleFromJsValue

Out of scope: handling null values in crazy places like inside of records and variants. Those aren't as high a priority so we will wait until someone reports an issue with those and fix them at that time.

lastmjs:
Implement Candid's `Null` type. I.e., you should now be able to write a canister with a method like this:

```typescript
export function someMethod(arg: null): Query<null> {
    return arg;
}
```

Resolves #200


dansteren (approved):
Looks good ;)

PR #371, opened 2022-06-09
Implement candid `reserved` type

### LINKED ISSUE #87: TypeScript -> Candid compiler implement reserved
~~Skipped for the time being, when it becomes important let's implement it. I think other things have a higher priority currently.~~

Implement Candid's `reserved` type. I.e., you should now be able to write a canister with a method like this:

```typescript
export function someMethod(arg: reserved): Query<reserved> {
    return arg;
}
```

Reserved is intended to be used to deprecate arguments. Because arguments are positional, you would use reserved to indicate that an early positional argument is no longer used but later ones still are. Unfortunately the candid spec indicates that the equivalent JS type is `any` hinting that the value could still be there. Additionally, you could put `reserved` as a tuple type or record sub-type or even as a return type, none of which make sense. So, for this issue we are treating `reserved` arguments as `undefined` but will allow you to return any value to Rust. It will just be ignored though, rather than attempting to convert it into some unknown type. In Motoko you might still be able to access those values, but we have consciously chosen to just clear it out instead.

Out of scope: handling `reserved` types within variants or records or other crazy places.

**Resources**
- Candid Reference: https://internetcomputer.org/docs/current/references/candid-ref/#type-reserved

dansteren:
This implementation adds support for the `reserved` candid type. The `reserved` type is aliased to `any` so that you can return whatever value you want from your canister. However, any `reserved` parameter will always be `null`. The `reserved` type loses all typing so Azle can't know what the value should be, so we always just leave it as null since it is intended to indicate that the value is no longer relevant to the method.

We considered making the reserved type `undefined` in the JS, but because the Rust and JS agents treat it as a null we did the same for uniformity.

Resolves #87


src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
```diff
@@ -167,5 +167,11 @@ export function generateAzleIntoJsValueTrait(): Rust {
                 boa_engine::JsValue::BigInt(boa_engine::bigint::JsBigInt::from_string(&self.0.to_string()).unwrap())
             }
         }
+
+        impl AzleIntoJsValue for ic_cdk::export::candid::Reserved {
+            fn azle_into_js_value(self, _: &mut boa_engine::Context) -> boa_engine::JsValue {
+                boa_engine::JsValue::Undefined
```
dansteren, line 173: Should this be `Null` instead of `Undefined`?

lastmjs: Let's change it to Null

examples/primitive_types/src/primitive_types.did
```diff
@@ -25,4 +25,7 @@ service: {
     "printFloat32": (float32) -> (float32) query;
     "getPrincipal": () -> (principal) query;
     "printPrincipal": (principal) -> (principal) query;
+    "getReserved": () -> (reserved) query;
+    "printReserved": (reserved) -> (reserved) query;
+    "skipReserved": (reserved, text) -> (text) query;
```
lastmjs, line 30: Remove

examples/primitive_types/test/test.ts
```diff
@@ -297,6 +297,42 @@ const tests: Test[] = [
...
+                Principal.fromText('rrkah-fqaaa-aaaaa-aaaaq-cai')
+            );
+
+            return {
+                ok: result === null
+            };
+        }
+    },
+    {
+        name: 'skipReserved',
```
lastmjs, line 324: Get rid of this

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
dansteren: Done

lastmjs (approved):

lastmjs (approved):

PR #372, opened 2022-06-10
Add candid empty type

### LINKED ISSUE #88: TypeScript -> Candid compiler implement empty
~~Skipped for the time being, when it becomes important let's implement it. I think other things have a higher priority currently.~~

Implement Candid's `empty` type. I.e., you should now be able to write a canister with a method like this:

```typescript
export function someMethod(arg: empty): Query<empty> {
    return arg;
}
```

The `empty` type is likely synonymous with void as it means no value at all. In practice JS devs should likely mark functions as returning `void` rather than `empty`, but this will cater to those wanting to be more in line with Candid's terminology.

Unfortunately, candid does not restrict you from using `empty` as a parameter type nor within variants nor records. As a result we need some way to represent this value in JS, for which `undefined` is likely the closest match. So, for the method above, the value of `arg` should be `undefined`. Meaning regardless of what is passed in, it will be coerced to `undefined`. Likewise, returning `undefined` is acceptable, though just returning (without a value) is likely more equivalent to the intended semantics (though the two are actually equivalent).

Out of scope: handling `empty` types within variants or records or other crazy places.

**Resources**
- Candid Reference: https://internetcomputer.org/docs/current/references/candid-ref/#type-empty

dansteren:
This implements candid's empty type..

Resolves #88

### DIFF
--- a/examples/primitive_types/src/primitive_types.ts
+++ b/examples/primitive_types/src/primitive_types.ts
@@ -1,5 +1,6 @@
 import {
     Query,
+    empty,
     int,
     int64,
     int32,
@@ -150,4 +151,15 @@ export function getReserved(): Query<reserved> {
 export function printReserved(reserved: reserved): Query<reserved> {
     console.log(typeof reserved);
     return reserved;
+}
+
+export function getEmpty(): Query<empty> {
+    throw 'Anything you want';
+}
+
+// Note: It is impossible to call this function because it requires an argument
+// but there is no way to pass an "empty" value as an argument.
+export function printEmpty(empty: empty): Query<empty> {
+    console.log(typeof empty);
+    throw 'Anything you want';
 }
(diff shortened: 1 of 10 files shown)


examples/primitive_types/src/primitive_types.ts
```diff
@@ -132,4 +133,15 @@ export function getPrincipal(): Query<Principal> {
...
     return principal;
-}
\ No newline at end of file
+}
+
+export function getEmpty(): Query<empty> {
+    throw 'Anything you want';
+}
+
+// Note: It is impossible to get into this function because it requires an
```
lastmjs, line 142: I suggest the word `call` instead of `get into` and `pass an "empty" value as an argument` instead of `pass an "empty" value.`

lastmjs (changes requested):

lastmjs (approved):

lastmjs (approved):

Issue #237, opened 2022-04-25
M1 Mac build issue - could not compile `wabt-sys`


krpeacock:
Issue can be observed in any example project, or even a blank ts entrypoint. Spec is 
OS: Monterey 12.3.1 
Processor: M1 Pro
Node: 16.14.0
Rust: 1.60.0

Full output is:

```
   Compiling wabt v0.10.0
error: failed to add native library /var/folders/xt/9tbb8ws541979mzdx2n9ms2r0000gn/T/cargo-installtXfSMj/release/build/wabt-sys-e78371680e273395/out/build/libwabt.a: file too small to be an archive

error: could not compile `wabt-sys` due to previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `ic-cdk-optimizer v0.3.4`, intermediate artifacts can be found at `/var/folders/xt/9tbb8ws541979mzdx2n9ms2r0000gn/T/cargo-installtXfSMj`
Caused by:
  build failed
Error: Command failed: cd target/azle && cargo install ic-cdk-optimizer --root ..
    at checkExecSyncError (node:child_process:828:11)
    at execSync (node:child_process:902:15)
    at installRustDependencies (/Users/kyle/code/azle-rss/backend/node_modules/azle/src/azle.ts:60:13)
    at azle (/Users/kyle/code/azle-rss/backend/node_modules/azle/src/azle.ts:26:5)
    at Object.<anonymous> (/Users/kyle/code/azle-rss/backend/node_modules/azle/src/azle.ts:15:1)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Module.m._compile (/Users/kyle/code/azle-rss/backend/node_modules/ts-node/src/index.ts:1365:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1155:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/kyle/code/azle-rss/backend/node_modules/ts-node/src/index.ts:1368:12)
    at Module.load (node:internal/modules/cjs/loader:981:32) {
  status: 101,
  signal: null,
  output: [ null, null, null ],
  pid: 97782,
  stdout: null,
  stderr: null
}
Error: The build step failed for canister 'rrkah-fqaaa-aaaaa-aaaaq-cai' with an embedded error: The custom tool failed.
```

lastmjs:
Related: https://forum.dfinity.org/t/setup-internet-identity-locally-failed-macbook-m1/7684

[...]

PR #419, opened 2022-06-16
change location of installation and execution for ic-cdk-optimizer to…

lastmjs:
… be compatible with Mac M1

Hopefully fixes: https://github.com/demergent-labs/azle/issues/237


dansteren (approved):
This is much better :)

PR #423, opened 2022-06-17
IC API `accept message`

### LINKED ISSUE #385: IC API accept_message
accept_message: https://docs.rs/ic-cdk/latest/ic_cdk/api/call/fn.accept_message.html

dansteren:
Implements the `accept_message` IC API call.

Resolves #385

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -52,6 +52,7 @@ ic.call_raw128 = function(...args) {
 };
 
 type ic = {
+    accept_message: () => void;
     // call: (
     //     canisterId: Principal,
     //     methodName: string,
--- a/src/compiler/typescript_to_rust/generators/ic_object/functions/accept_message.ts
+++ b/src/compiler/typescript_to_rust/generators/ic_object/functions/accept_message.ts
@@ -1,0 +1,13 @@
+import { Rust } from '../../../../../types';
+
+export function generateIcObjectFunctionAcceptMessage(): Rust {
+    return /* rust */ `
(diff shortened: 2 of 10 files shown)


index.ts
```diff
@@ -52,6 +52,7 @@ ic.call_raw128 = function(...args) {
 };
 
 type ic = {
+    accept_message: () => void;
```
lastmjs, line 55: Should this be void or undefined?

src/compiler/typescript_to_rust/generators/ic_object/functions/accept_message.ts
```diff
@@ -0,0 +1,13 @@
+import { Rust } from '../../../../../types';
+
+export function generateIcObjectFunctionAcceptMessage(): Rust {
+    return /* rust */ `
+        fn _azle_ic_accept_message(
+            _this: &boa_engine::JsValue,
+            _aargs: &[boa_engine::JsValue],
+            _context: &mut boa_engine::Context
+        ) -> boa_engine::JsResult<boa_engine::JsValue> {
+            Ok(ic_cdk::api::call::accept_message().azle_into_js_value(_context))
```
lastmjs, line 10: This will create a JsValue::Undefined correct? I wonder if we should then make the TypeScript type return undefined

lastmjs (changes requested):

lastmjs (approved):

PR #447, opened 2022-06-25
Implement `reject_code`

### LINKED ISSUE #393: IC API reject_code
reject_code: https://docs.rs/ic-cdk/latest/ic_cdk/api/call/fn.reject_code.html

dansteren:
Resolves #393

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -87,6 +87,7 @@ type ic = {
     msg_cycles_refunded128: () => nat;
     print: (...args: any) => void;
     reject: (message: string) => void;
+    reject_code: () => RejectionCode;
     stableStorage: <T>() => T;
     time: () => nat64;
     trap: (message: string) => never;
@@ -185,3 +186,13 @@ export type Func<
 > = [Principal, string];
 
 export { Principal } from '@dfinity/principal';
+
+export type RejectionCode = Variant<{
+    NoError: null;
+    SysFatal: null;
+    SysTransient: null;
+    DestinationInvalid: null;
+    CanisterReject: null;
+    CanisterError: null;
+    Unknown: null;
+}>;
--- a/src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
+++ b/src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
@@ -64,6 +64,83 @@ export function generateAzleIntoJsValueTrait(): Rust {
             }
         }
 
+        impl AzleIntoJsValue for ic_cdk::api::call::RejectionCode {
+            fn azle_into_js_value(self, context: &mut boa_engine::Context) -> boa_engine::JsValue {
+                match self {
+                    ic_cdk::api::call::RejectionCode::NoError => {
+                        boa_engine::object::ObjectInitializer::new(context)
+                            .property(
+                                "NoError",
+                                boa_engine::JsValue::Null,
+                                boa_engine::property::Attribute::all(),
+                            )
+                            .build()
+                            .into()
+                    }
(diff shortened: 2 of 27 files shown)


index.ts
```diff
@@ -182,3 +184,13 @@ export type Func<
 > = [Principal, string];
 
 export { Principal } from '@dfinity/principal';
+
+export type RejectionCode = Variant<{
+    NoError?: null;
```
lastmjs, line 189: No need for the `?`s

examples/responses/canisters/rejecter/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Canister, CanisterResult, ic, Principal } from 'azle';
+
+export type Rejector = Canister<{
```
lastmjs, line 3: You're using two different spelling of rejecter/rejector

examples/responses/test/test.ts
```diff
@@ -0,0 +1,30 @@
...
+        host: 'http://127.0.0.1:8000'
+    }
+});
+
+const tests: Test[] = [
+    ...cleanDeploy('responses'),
+    {
+        name: 'reject codes',
+        test: async () => {
+            // TODO: WRITE TESTS!!!
```
lastmjs, line 22: You didn't write tests (just kidding, but not, but still just kidding)

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
```diff
@@ -64,6 +64,83 @@ export function generateAzleIntoJsValueTrait(): Rust {
             }
         }
 
+        impl AzleIntoJsValue for ic_cdk::api::call::RejectionCode {
```
lastmjs, line 67: This is pretty impressive!

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
```diff
@@ -64,6 +64,83 @@ export function generateAzleIntoJsValueTrait(): Rust {
...
+                    ic_cdk::api::call::RejectionCode::NoError => {
+                        boa_engine::object::ObjectInitializer::new(context)
+                            .property(
+                                "NoError",
+                                boa_engine::JsValue::Null,
+                                boa_engine::property::Attribute::all(),
+                            )
+                            .build()
+                            .into()
+                    }
```
lastmjs, line 79: I don't understand the syntax here, I thought a comma would be necessary

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
lastmjs, line 67: We might want to write a Vec impl for ic_cdk::api::call::RejectionCode as well just in case and since all of the other types have Vec impls

lastmjs (changes requested):

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
dansteren: Commas between `.build()` and `.into()`? This is a boa ObjectInitializer. From what I'm seeing in [the docs](https://docs.rs/boa_engine/0.15.0/boa_engine/object/struct.ObjectInitializer.html) and other places in the codebase you call `::new` on it to generate a new builder thing, then call `.property` to add properties, and finish it up with a `.build` to convert it from an ObjectInitializer to an Object. The `.into` just makes it a JsValue. So, commas within calls to `.property` but not between the other calls.

Happy to talk in person if I'm just not understanding the question or something.

PR #455, opened 2022-06-30
change getCount from Update to Query

bdemann:


dansteren:
Can you manually run prettier on your files (`npx prettier --write .`)? I think that's where the majority of the changes in this PR are from. I think even though the files are auto-generated by `dfx generate` we are still running them through prettier. So each time you run `dfx generate` run prettier after. We can work on getting it in a pre-commit hook. Just haven't gotten it that far yet.

examples/motoko_examples/minimal-counter-dapp/src/declarations/minimal_dapp/index.js
```diff
@@ -1,4 +1,4 @@
-import { Actor, HttpAgent } from '@dfinity/agent';
+import { Actor, HttpAgent } from "@dfinity/agent";
```
lastmjs, line 1: First step, if you could run `npx prettier --write .`, that will clean up a lot of these files.

lastmjs (changes requested):

examples/motoko_examples/minimal-counter-dapp/src/declarations/minimal_dapp/index.js
lastmjs: Then incorporate the latest from main and push then this should be good

lastmjs (approved):

PR #468, opened 2022-07-02
Implement IC API `reply`

### LINKED ISSUE #395: IC API reply
reply: https://docs.rs/ic-cdk/latest/ic_cdk/api/call/fn.reply.html

- [ ] Check out the `_this` parameter on Rust to JS FFIs to see if the top-level calling function name is there, if it is we can just expose one `ic.reply` function and match on the specific typed reply Rust function to call it and have the Candid ser/der done for us correctly
- [x] Create special `reply` functions on the ic object

dansteren:
Resolves #395


examples/manual_reply/test/test.ts
```diff
@@ -69,18 +65,110 @@ const tests: Test[] = [
...
+            };
+        }
+    },
+    {
+        name: 'reply with float32',
+        test: async () => {
+            const result = await manual_reply_canister.reply_float32();
+
+            return {
+                ok: result === 1245.677978515625 // TODO: This should not be changing
```
lastmjs, line 94: What's wrong here?

lastmjs (changes requested):

examples/manual_reply/test/test.ts
dansteren: The canister returns `1245.678` in the typescript code but when we get outside the canister we have `1245.677978515625` instead. It shouldn't lose it's precision like that.

examples/manual_reply/test/test.ts
```diff
@@ -10,77 +12,295 @@ const manual_reply_canister = createActor('rrkah-fqaaa-aaaaa-aaaaq-cai', {
...
+        }
+    },
+    {
+        name: 'update reply with float32',
+        test: async () => {
+            // const result = await manual_reply_canister.update_float32();
+            // Note: The JS agent doesn't handle floats correctly.
+            // See https://github.com/dfinity/agent-js/issues/589
+            const result = execSync(
+                'dfx canister call manual_reply query_float32'
```
lastmjs, line 65: Shouldn't this be a call to `update_float32`?

src/compiler/typescript_to_rust/ast_utilities/miscellaneous.ts
```diff
@@ -48,7 +48,14 @@ export function getRustTypeNameFromTypeNode(typeNode: tsc.TypeNode): Rust {
...
     }
 
-    // TODO possibly type literals?
+    if (typeNode.kind === tsc.SyntaxKind.LiteralType) {
+        const literalTypeNode = typeNode as tsc.LiteralTypeNode;
+
+        if (literalTypeNode.literal.kind === tsc.SyntaxKind.NullKeyword) {
+            return `()`;
+        }
+        // TODO possibly other type literals?
```
lastmjs, line 57: Could you put this comment outside of this if statement, and/or make another comment to explain that we need support for type literals and literal types? Type literals and literal types I believe are different concepts. I believe `tsc.SyntaxKind.LiteralType` and `tsc.SyntaxKind.TypeLiteral` are not the same thing, I assume a `TypeLiteral` is like an object literal, `{}` with properties. LiteralTypes might be things like null (in this case).

lastmjs (changes requested):

examples/manual_reply/test/test.ts
dansteren: Good catch. This was definitely just copy/pasta so good eye. I've pushed a change.

src/compiler/typescript_to_rust/ast_utilities/miscellaneous.ts
dansteren: I've made a second comment.

lastmjs (dismissed):

lastmjs:
If you get this up to date then it's ready

PR #469, opened 2022-07-06
381 ic api stable grow

bdemann:


index.ts
```diff
@@ -206,3 +209,13 @@ export type RejectionCode = Variant<{
...
+
+export type StableMemoryError = Variant<{
+    OutOfMemory: null;
+    OutOfBounds: null;
+}>;
+
+export type StableGrowResult = Variant<{
+    ok: nat32;
+    err: StableMemoryError;
+}>;
```
dansteren, line 221: @lastmjs just for learning's sake, why can't we do `Result<nat32, StableMemoryError>`? I see that you've commented out those types so I'm assuming they don't work for some reason. Otherwise that seems more elegant to me.

examples/stable_memory/test/test.ts
```diff
@@ -0,0 +1,78 @@
...
+            const new_pages = 5;
+            const result = await stable_memory_canister.stable_grow(new_pages);
+            const new_size = await stable_memory_canister.stable_size();
+
+            return {
+                ok:
+                    'ok' in result &&
+                    result.ok === old_size &&
+                    new_pages + old_size === new_size
+            };
```
dansteren, line 45: @bdemann Just in case you hadn't see it before, you could explicitly handle the err case of your result first:

```ts
import { ok } from 'azle'

if(!ok(result)) {
  return { err: result.err };
}
```
Then afterwards you'd still have to check `result.ok === old_size && new_pages + old_size === new_size`, but the advantage would be that if there was an error, the testing framework would print out the err message, vs with that you have the test would just fail, but wouldn't show what the error message was.

examples/stable_memory/test/test.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+    },
+    {
+        name: 'stable grow out of memory',
```
dansteren, line 66: I'm not sure how big a deal this is, but this test is dependent upon the state of the last one. If we were using a more robust framework I would suggesting clearing the memory between each test or something and then doing all your setup for this test within this function. That way it won't break if we add another test or move this one above the "stable grow to max" test in the future.

dansteren (approved):
I left a couple of suggestions but think this PR is also good as is.

examples/stable_memory/test/test.ts
bdemann: My results are ```StableGrowResults``` that don't seem to work with ```ok```
The best I was able to come up with is 
```
            if ('err' in result) {
                return {
                    err: Object.getOwnPropertyNames(result.err)[0]
                };
            }
```
Which seems a little clunky to me. What are you thoughts? Or is there a way to make StableGrowResults work with ok?

examples/stable_memory/test/test.ts
bdemann: Do we have an easy way to clear the memory? The only way I know how to do it is to uninstall the canister code.

src/compiler/typescript_to_rust/generators/ic_object/functions/stable_grow.ts
```diff
@@ -0,0 +1,14 @@
+import { Rust } from '../../../../../types';
+
+export function generateIcObjectFunctionStableGrow(): Rust {
+    return /* rust */ `
+        fn _azle_ic_stable_grow(
+            _this: &boa_engine::JsValue,
+            _aargs: &[boa_engine::JsValue],
+            _context: &mut boa_engine::Context
+        ) -> boa_engine::JsResult<boa_engine::JsValue> {
+            let new_pages: u32 = _aargs.get(0).unwrap().clone().azle_try_from_js_value(_context).unwrap();
```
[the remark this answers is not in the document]
bdemann: Oh, is this not something that prettier takes care of? I was just assuming that it was as pretty as it could be

examples/stable_memory/test/test.ts
bdemann: @lastmjs said that we should be okay to do this for these tests since they are smoke tests. For the property tests we will want to be much more exact.

index.ts
bdemann: @lastmjs says he doesn't remember why it's commented out, its a valid concern and we will address it later if we reintroduce the Result type, but for right now it's okay

lastmjs (approved):

PR #473, opened 2022-07-06
384 ic api stable write

bdemann:


examples/stable_memory/src/stable_memory.ts
```diff
@@ -23,3 +24,7 @@ export function stable_grow(new_pages: nat32): Update<StableGrowResult> {
 export function stable64_grow(new_pages: nat64): Update<Stable64GrowResult> {
     return ic.stable64_grow(new_pages);
 }
+
+export function stable_write(offset: nat32, buffer: nat8[]): Update<void> {
```
lastmjs: blob instead of nat8[]

examples/stable_memory/test/test.ts
```diff
@@ -77,6 +77,49 @@ const tests: Test[] = [
...
+            await stable_memory_canister.stable_write(offset, buffer);
+
+            return {
+                ok: true
+            };
+        }
+    },
+    {
+        name: 'stable write out of bounds error',
+        test: async () => {
```
lastmjs: At least this test should be complete

lastmjs (approved):

PR #476, opened 2022-07-06
Bump IC CDK to v0.5.2

### LINKED ISSUE #471: Update `ic_cdk` crate dependency to > 0.5.2
We need a new version of the IC CDK to be able to implement some features such as #462. At time of writing the latest version of that crate is 0.5.2 but we should update to the latest version when we get to this card.

dansteren:
Resolves #471


lastmjs:
I'm curious why this pull request is here? I was thinking it would just get updated in our pull requests (since we've each done this)

PR #479, opened 2022-07-07
Implement IC API `arg data raw`

### LINKED ISSUE #462: IC API arg_data_raw
Expose the `arg_data_raw` IC API on the `ic` object in js. This should allow you do the following:

```ts
export function get_arg_data_raw(arg1: string, arg2: nat, etc: any): Query<blob> {
  return ic.arg_raw_data()
}
```
See https://docs.rs/ic-cdk/latest/ic_cdk/api/call/fn.arg_data_raw.html

dansteren:
Resolves #462.

### DIFF
--- a/examples/ic_api/src/ic_api.ts
+++ b/examples/ic_api/src/ic_api.ts
@@ -2,6 +2,7 @@ import {
     blob,
     empty,
     ic,
+    int8,
     nat,
     nat64,
     Opt,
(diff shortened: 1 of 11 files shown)


examples/ic_api/src/ic_api.ts
```diff
@@ -11,6 +12,16 @@ import {
     Update
 } from 'azle';
 
+// returns the argument data as bytes.
+export function arg_data_raw(
```
lastmjs, line 16: Let's add one record and one enum as well

lastmjs (changes requested):

examples/ic_api/src/ic_api.ts
dansteren: There are issues with records and variants. I've made https://github.com/demergent-labs/azle/issues/487 to track it.

lastmjs (approved):

Issue #470, opened 2022-07-06
Add JS Docs for azle


dansteren:
**As an** Azle canister developer,
**I want** inline documentation when I hover over an import from Azle,
**So that** I don't have to go look at separate documentation.

Current Behavior: Hovering over an import from Azle provides only the type signature, not the JS doc info:
![image](https://user-images.githubusercontent.com/5455419/177490710-7d973020-d27e-40f2-b41d-7c325493c60b.png)

Expected behavior: Hovering over an import from Azle provides informative documentation explaining what it does, params, return types etc:
![image](https://user-images.githubusercontent.com/5455419/177490988-efe4ed96-6fe7-411a-ae35-f12811d68fb5.png)

This might be out of scope for this ticket, but we should consider using [TypeDoc](https://typedoc.org/) or another tool to publish these as well.

PR #480, opened 2022-07-07
Implement IC API `arg_data_raw_size`

### LINKED ISSUE #463: IC API arg_data_raw_size
Expose the `arg_data_raw_size` IC API on the `ic` object in ts. This should allow you do the following:

```ts
export function get_arg_data_raw_size(arg1: string, arg2: nat, etc: any): Query<nat> {
  return ic.arg_raw_data_size()
}
```
> :warning: make sure to get the correct candid equivalent of `usize`

See https://docs.rs/ic-cdk/latest/ic_cdk/api/call/fn.arg_data_raw_size.html

dansteren:
Resolves #463.

### DIFF
--- a/examples/ic_api/src/ic_api.ts
+++ b/examples/ic_api/src/ic_api.ts
@@ -4,7 +4,7 @@ import {
     ic,
     int8,
     nat,
-    nat64,
+    nat32,
     Opt,
     Principal,
     Query,
@@ -20,6 +20,16 @@ export function arg_data_raw(
     arg4: string
 ): Query<blob> {
     return ic.arg_data_raw();
+}
+
+// returns the length of the argument data in bytes
+export function arg_data_raw_size(
+    arg1: blob,
+    arg2: int8,
+    arg3: boolean,
+    arg4: string
+): Query<nat32> {
+    return ic.arg_data_raw_size();
 }
 
 // returns the principal of the identity that called this function
--- a/examples/ic_api/test/test.ts
+++ b/examples/ic_api/test/test.ts
@@ -34,6 +34,35 @@ const tests: Test[] = [
         }
     },
     {
+        name: 'arg_data_raw_size',
+        test: async () => {
+            const blobString = 'Surprise!';
+            const blob = blobString.split('').map((char) => char.charCodeAt(0));
+            const int = 127;
+            const bool = true;
+            const string = 'test';
+
+            const resultArgDataRawSize =
+                await ic_api_canister.arg_data_raw_size(
+                    blob,
+                    int,
+                    bool,
+                    string
+                );
+
+            const resultArgDataRaw = await ic_api_canister.arg_data_raw(
+                blob,
+                int,
+                bool,
+                string
+            );
+
+            return {
+                ok: resultArgDataRawSize === resultArgDataRaw.length
+            };
+        }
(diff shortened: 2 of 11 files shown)


examples/ic_api/src/ic_api.ts
```diff
@@ -11,6 +12,26 @@ import {
...
+    arg1: blob,
+    arg2: int8,
+    arg3: boolean,
+    arg4: string
+): Query<blob> {
+    return ic.arg_data_raw();
+}
+
+// returns the length of the argument data in bytes
+export function arg_data_raw_size(
```
lastmjs, line 26: Let's add one record and one enum

examples/ic_api/test/test.ts
```diff
@@ -10,6 +10,51 @@ const ic_api_canister = createActor('rrkah-fqaaa-aaaaa-aaaaq-cai', {
...
+
+            const result = await ic_api_canister.arg_data_raw_size(
+                blob,
+                int,
+                bool,
+                string
+            );
+
+            return {
+                ok: result === 29n
```
lastmjs, line 54: I think we should get the length as part of the test instead of hard-coding it, doing a call to ic_api_canister.arg_data and getting the length

index.ts
```diff
@@ -53,6 +53,12 @@ ic.call_raw128 = function (...args) {
 
 type ic = {
     accept_message: () => void;
+    /**
+     * Returns the argument data as bytes.
+     * @returns the argument data
+     */
+    arg_data_raw: () => blob;
+    arg_data_raw_size: () => nat64;
```
lastmjs, line 61: Why no JS/TS doc here?

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
```diff
@@ -210,6 +210,12 @@ export function generateAzleIntoJsValueTrait(): Rust {
             }
         }
 
+        impl AzleIntoJsValue for usize {
+            fn azle_into_js_value(self, context: &mut boa_engine::Context) -> boa_engine::JsValue {
+                boa_engine::JsValue::BigInt(self.into())
```
lastmjs, line 215: I'm not sure this should be a BigInt, let's discuss

lastmjs (changes requested):

examples/ic_api/src/ic_api.ts
dansteren: There are issues with records and enums. You can track this in https://github.com/demergent-labs/azle/issues/487.

index.ts
dansteren: I added it for arg_data_raw so we could have a discussion. I've removed it though and will wait until we do #470 to add them in consistenly.

src/compiler/typescript_to_rust/generators/azle_into_js_value_trait.ts
dansteren: I've changed this to nat32.

examples/ic_api/test/test.ts
dansteren: Agreed. I've fixed this. Thanks for the help!

lastmjs (approved):

PR #494, opened 2022-07-11
fix null type parsing and add null tests

### LINKED ISSUE #475: null type everywhere
I've run into a few problems with using the `null` type, make sure it can be used in parameters, return types, records, and variants.

bdemann:
Fixes #475

### DIFF
--- a/examples/null_example/src/null_example.ts
+++ b/examples/null_example/src/null_example.ts
@@ -1,0 +1,63 @@
+import { int, Query, Update, Variant } from 'azle';
+
+type PartiallyNullRecord = {
+    first_item: int;
+    second_item: null;
+    third_item: int;
+};
+
+type TwoNullRecord = {
+    first_item: null;
+    second_item: null;
+};
+
+type ThreeNullRecord = {
+    first_item: null;
+    second_item: null;
+    third_item: null;
+};
+
+export function null_function(param: null): Query<null> {
+    return param;
+}
+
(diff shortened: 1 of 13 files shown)


examples/null_example/src/null_example.ts
```diff
@@ -0,0 +1,63 @@
+import { int, Query, Update, Variant } from 'azle';
+
+type partially_null_object = {
```
lastmjs, line 3: The convention for type names is to have them by uppercase camelcase, so this would be `PartiallyNullObject`

examples/null_example/src/null_example.ts
```diff
@@ -0,0 +1,63 @@
...
+    first_item: null;
+    second_item: null;
+    third_item: null;
+};
+
+export function null_function(param: null): Query<null> {
+    return param;
+}
+
+export function get_partial_null_record(): Query<partially_null_object> {
```
lastmjs, line 24: If we're calling this `get_partial_null_record` then I think we should name the type `PartiallyNullRecord`

lastmjs (changes requested):
Just a couple name changes

examples/null_example/src/null_example.ts
bdemann: Sounds good, and I'll change the function to ```get_partially_null_record``` while I'm at it

lastmjs (dismissed):

PR #495, opened 2022-07-11
Remove `ic.arg_data` and add rough example for `ic.result`

dansteren:
The `ic.arg_data` API wasn't implemented correctly so this removes it until we can get it fully working.

Additionally, this start the implementation of `ic.result` but because it depends on `ic.arg_data` it is only roughly implemented and all commented out.

All of this to be fully implemented by #496.

### DIFF
--- a/examples/rejections/test/test.ts
+++ b/examples/rejections/test/test.ts
@@ -61,6 +61,7 @@ const tests: Test[] = [
         }
     },
     {
+        skip: true,
         name: 'result with an accept',
         test: async () => {
             const rejectionMessage = 'custom rejection message';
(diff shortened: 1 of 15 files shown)


examples/rejections/test/test.ts
```diff
@@ -59,6 +59,46 @@ const tests: Test[] = [
                 ok: result === rejectionMessage
             };
         }
+    },
+    {
+        name: 'result with an accept',
```
lastmjs, line 64: Skip these tests

lastmjs (changes requested):

PR #507, opened 2022-07-12
added motoko and rust examples to query example

bdemann:


bdemann:
Note: since this example only has queries no benchmarks were added

examples/query/canisters/motoko/main.mo
```diff
@@ -0,0 +1,5 @@
+actor {
+  public func simple_query() : async Text {
```
lastmjs, line 2: I don't believe this is a query function

examples/motoko_examples/calc/src/calc.did
```diff
@@ -4,4 +4,4 @@ service: {
     "mul": (int) -> (int);
     "div": (int) -> (opt int);
     "clearall": () -> ();
-}
+}
```
lastmjs, line 7: We should have newlines at the end of our files, @dansteren can help you set up VS Code automation for this if you want

examples/query/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[workspace]
+members = [
+    "canisters/rust",
+]
+
+[profile.release]
+lto = true
+opt-level = 'z'
+codegen-units = 1
```
lastmjs, line 9: We should have newlines at the end of our files, @dansteren can help you set up VS Code automation for this if you want. This comment goes for all files in this pull request

examples/query/canisters/rust/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[ic_cdk_macros::query]
+fn simple_query() -> String {
+    format!("This is a query function")
```
lastmjs, line 3: Use `"This is a query function".to_string()`, format! is a macro with special capabilities, I would say overkill for just returning a string. The convention used in most of the other Rust codebases that I have use .to_string or String::from()

examples/query/canisters/rust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "rust"
+version = "0.0.0"
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+candid = "0.7.14"
```
lastmjs, line 10: Do you need the `candid` and `ic-cdk` dependencies here? Let's see if we can remove them

examples/motoko_examples/calc/src/calc.did
lastmjs: Also why is this file even in the pull request?

lastmjs (changes requested):

lastmjs:
Also, it doesn't seem like prettier is running on our Motoko code. @dansteren can you help us out here?

examples/query/canisters/rust/Cargo.toml
bdemann: When I got rid of ic-cdk I got this error
```
--> canisters/rust/src/lib.rs:1:1
  |
1 | #[ic_cdk_macros::query]
  | ^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `ic_cdk`
  |
  = note: this error originates in the attribute macro `ic_cdk_macros::query` (in Nightly builds, run with -Z macro-backtrace for more info)
```

dansteren:
> Also, it doesn't seem like prettier is running on our Motoko code. @dansteren can you help us out here?

Prettier doesn't/can't run on motoko code. Prettier parses the code into an AST and then re-writes it out to the file using whatever standards it decides. But so in order for it to work on Motoko, they would have to understand the AST of motoko code, which it currently doesn't. I think there might be people working on this, but there isn't a solution right now.

lastmjs (dismissed):

PR #520, opened 2022-07-13
Implemented benchmarking with automatic GitHub Action benchmarking process

lastmjs:


lastmjs:
The current issue with failing benchmarks is that the branch needs to be up-to-date

PR #523, opened 2022-07-14
Added benchmarks for examples/motoko_examples/quicksort

### LINKED ISSUE #518: examples/motoko_examples/quicksort benchmarks
Re-write the example at examples/motoko_examples/quicksort in Rust/Motoko and generate corresponding benchmarks.

- [x] Rust
- [x] Motoko
- [x] Azle
- [x] Benchmarks

dansteren:
This adds a Rust and Motoko implementation of this example and benchmarks for all 3 implementations.

Resolves #518


examples/motoko_examples/quicksort/Cargo.toml
```diff
@@ -0,0 +1,4 @@
+[workspace]
```
lastmjs, line 1: Should have the optimization stuff

examples/motoko_examples/quicksort/canisters/rust/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+mod quicksort;
+
+#[ic_cdk_macros::query]
+fn sort(xs: Vec<i128>) -> Vec<i128> {
```
lastmjs, line 4: Let's consider using `ic_cdk::export::candid::Int`

examples/motoko_examples/quicksort/dfx.json
```diff
@@ -1,14 +1,28 @@
 {
     "canisters": {
-        "quicksort": {
+        "azle": {
```
lastmjs, line 3: Same issue is in the other pull requests, dfx.json should look like the others

examples/motoko_examples/quicksort/dfx_generated/motoko/index.js
```diff
@@ -0,0 +1,40 @@
...
+        canisterId,
+        ...options?.actorOptions
+    });
+};
+
+/**
+ * A ready-to-use agent for the motoko canister
+ * @type {import("@dfinity/agent").ActorSubclass<import("./motoko.did.js")._SERVICE>}
+ */
+export const motoko = createActor(canisterId);
```
lastmjs, line 40: Not commented out

examples/motoko_examples/quicksort/canisters/motoko/Quicksort.mo
```diff
@@ -0,0 +1,55 @@
+import Array "mo:base/Array";
+import Order "mo:base/Order";
+import Int "mo:base/Int";
+
+module Quicksort {
+
+  type Order = Order.Order;
+
+  // Sort the elements of an array using the given comparison function.
+  public func sortBy<X>(xs : [X], f : (X, X) -> Order) : [X] {
```
lastmjs, line 10: The Motoko and Rust algorithms are not identical, there are various differences

lastmjs (changes requested):

examples/motoko_examples/quicksort/canisters/rust/src/lib.rs
dansteren: The example has been updated to use Ints. Marking this as resolved.

examples/motoko_examples/quicksort/canisters/motoko/Quicksort.mo
dansteren: I've done my best to align the two algorithms. There's one small difference with how I do the swap because in rust you need two mutable references to the same array. Doing that myself would require an unsafe block so I just used the builtin swap method. I think it's essentially equivalent under the hood, it just doesn't look that way at first glance.

TLDR; I think they're as close as they can get now. I'll let you evaluate that though and mark this thread resolved if you feel confident.

examples/motoko_examples/quicksort/canisters/rust/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+mod quicksort;
+use ic_cdk::export::candid::Int;
+
+#[ic_cdk_macros::query]
```
lastmjs, line 4: I didn't realize these didn't have benchmarks, I think because of the usefulness of these benchmarks that we should change them to update functions and benchmark them.

examples/motoko_examples/quicksort/canisters/rust/src/quicksort.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+fn sort_by_helper<X: Clone + Ord, T: Fn(&X, &X) -> Ordering>(
+    xs: &mut Vec<X>,
+    l: Int,
+    r: Int,
+    f: &T,
+) -> () {
+    if l < r {
+        let mut i = l.clone();
```
lastmjs, line 37: Let's just see if we can get rid of all of these clones

lastmjs (changes requested):

examples/motoko_examples/quicksort/canisters/rust/src/quicksort.rs
dansteren: I don't think that's possible. If these were on the stack then assigning them to multiple variables would be no problem because it would be a cheap copy. But because Int is stored on the heap, setting `i` to `l` would actually be a `move` which means `l` wouldn't be available anymore. Thus, the copy. If I'm mistaken or there's a way to do this I'm happy to work through this with you.

lastmjs (approved):

PR #546, opened 2022-07-19
Fix prettier pre-commit hook

dansteren:
Prettier was only running on some supported file types. Now it runs on all and ignores any known types.


lastmjs:
Does this PR include the results of running it?

dansteren:
> Does this PR include the results of running it?

I just ran it locally and it didn't make any changes so all files should be good at this point.

PR #547, opened 2022-07-19
545 notify raw

### LINKED ISSUE #545: notify_raw
For some reason we didn't implement this.

lastmjs:
closes #545

### DIFF
--- a/examples/notify_raw/test/test.ts
+++ b/examples/notify_raw/test/test.ts
@@ -1,0 +1,60 @@
+import { large_wasm_deploy, ok, run_tests, Test } from 'azle/test';
+import { createActor as createActorCanister1 } from './dfx_generated/canister1';
+import { createActor as createActorCanister2 } from './dfx_generated/canister2';
+
+const canister1 = createActorCanister1('rrkah-fqaaa-aaaaa-aaaaq-cai', {
+    agentOptions: {
+        host: 'http://127.0.0.1:8000'
+    }
+});
+
+const canister2 = createActorCanister2('ryjl3-tyaaa-aaaaa-aaaba-cai', {
+    agentOptions: {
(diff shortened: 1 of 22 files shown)


examples/notify_raw/test/test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const canister2 = createActorCanister2('ryjl3-tyaaa-aaaaa-aaaba-cai', {
+    agentOptions: {
+        host: 'http://127.0.0.1:8000'
+    }
+});
+
+const tests: Test[] = [
+    ...large_wasm_deploy('canister1'),
+    ...large_wasm_deploy('canister2'),
```
dansteren, line 19: You'll need to use the new `deploy` function when you merge main in.

index.ts
```diff
@@ -122,6 +122,12 @@ type ic = {
     msg_cycles_available128: () => nat;
     msg_cycles_refunded: () => nat64;
     msg_cycles_refunded128: () => nat;
+    notify_raw: (
+        canister_id: Principal,
+        method: string,
+        args_raw: blob,
+        payment: nat
```
dansteren, line 129: Is this payment the same as our `.with_cycles` flow that we use elsewhere? If so, should we not use that instead? If not, what is this payment?

dansteren (changes requested):
Everything looks good, other than the change to `cleanDeploy`. But because I have questions on the payment flow I'm requesting changes for now... But let's review and I'm happy to approve this once we talk through the payment stuff.

index.ts
lastmjs: It's not the same, it's entirely separate. This is like call_raw, you just call it like a function and pass in a payment if you wish

bdemann (changes requested):
Please add the tests to the test.yaml

PR #550, opened 2022-07-20
Added Rust and Motoko examples to echo

bdemann:


examples/motoko_examples/echo/dfx_generated/motoko/index.js
```diff
@@ -0,0 +1,40 @@
...
+        canisterId,
+        ...options?.actorOptions
+    });
+};
+
+/**
+ * A ready-to-use agent for the motoko canister
+ * @type {import("@dfinity/agent").ActorSubclass<import("./motoko.did.js")._SERVICE>}
+ */
+export const motoko = createActor(canisterId);
```
dansteren, line 40: We're going to need this line commented out to be able to use it in our tests or benchmarks.

examples/motoko_examples/echo/dfx_generated/azle/index.js
```diff
@@ -34,7 +34,7 @@ export const createActor = (canisterId, options) => {
 };
 
 /**
- * A ready-to-use agent for the echo canister
- * @type {import("@dfinity/agent").ActorSubclass<import("./echo.did.js")._SERVICE>}
+ * A ready-to-use agent for the azle canister
+ * @type {import("@dfinity/agent").ActorSubclass<import("./azle.did.js")._SERVICE>}
  */
-//  export const echo = createActor(canisterId);
+export const azle = createActor(canisterId);
```
dansteren, line 40: Comment out this last line.

examples/motoko_examples/echo/dfx_generated/rust/index.js
```diff
@@ -0,0 +1,40 @@
...
+        canisterId,
+        ...options?.actorOptions
+    });
+};
+
+/**
+ * A ready-to-use agent for the rust canister
+ * @type {import("@dfinity/agent").ActorSubclass<import("./rust.did.js")._SERVICE>}
+ */
+export const rust = createActor(canisterId);
```
dansteren, line 40: Comment out this line.

dansteren (changes requested):
Looks really good. You just need to comment out the last lines in the agent files so that we can use them in our tests/benchmarks.

dansteren:
The other thing I noticed is that some of your files don't have a final newline character. We should be adding final newlines to all files since that's how [POSIX defines a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206). You can see this in your PRs where the files have a :no_entry: emoji at the very end. Fortunately, this is a really easy problem to solve. Just enable the "Files: Insert Final Newline" setting in VS Code:
image
While you're at it I would also recommend you enable "Files: Trim Final Newlines" and "Files: Trim Trailing Whitespace".

Ideally we'd have a `.editorconfig` in our repo that would ensure this happens but manually setting it for now should be sufficient.

lastmjs (approved):

PR #552, opened 2022-07-20
Create canister first in tests

dansteren:
If the canister doesn't exist, attempting to clear the memory first
results in DFX not being able to determine the ID from the canister
name. This way the canister will always exist before we attempt the
other operations.
Incorrect deploy order


dansteren:
**TLDR;** if it doesn't matter either way, this would make my life easier.

---

Although there is an "error" either way, this order feels much better to me. The current setup actually says "error" in my terminal in red (which catches my eye and makes me think there's something wrong):

image

Whereas if we switch to create the canister first, sure it "errors" but it isn't in red, and it just says the canister already exists

image
In fact, it's still an exit code of 0:
image
But uninstalling code from a non-existant canister isn't a 0 exit code:
image

lastmjs (approved):

PR #572, opened 2022-07-28
Did some tweaks to the main benchmarking code

lastmjs:
a few tweaks to the main benchmarking code, run the bytes and primitive_ops benchmarks


dansteren:
Can you please add a trailing newline to the benchmark generation?

dansteren (approved):
Resounding yes!

dansteren:
If we're being particular about PR names to make releases easier, I would suggest renaming this PR to: "Changed benchmark output format" or something.

I think it'd be great to standardize around using the keywords from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/#how).

PR #582, opened 2022-08-03
Implemented outgoing http requests with a couple examples

### LINKED ISSUE #566: Implement HTTP functionality
- [x] Provide types
- [x] Create examples
- [x] Implement Chainlink example with quicknodes (make this a separate repo, probably not tests necessary)
- [x] Create a frontend to show people that this works?? I couldn't deploy for a few weeks anyway...might be nice to have if it's really simple
- [x] Get curl commands for Chainlink price feeds, document them
- [x] Attempt to get one Ethereum JSON RPC endpoint to work from a canister
- [x] Add documentation
- [x] Add to the management canister example
- [x] Cycle costs: (it's about $0.004 per fetch, about $15 per month including the heartbeat every 10 minutes)

lastmjs:
Closes #566


examples/ethereum_json_rpc/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        yield ManagementCanister.http_request({
+            url: stable_storage.ethereum_url,
+            max_response_bytes,
+            http_method: {
+                POST: null
+            },
+            headers: [],
+            body: new Uint8Array(
+                encodeUtf8(
+                    JSON.stringify({
```
dansteren, line 46: If it were me, I would abstract new `Uint8Array(encodeUtf8(JSON.stringify({}))` into some method like `createHttpRequestBody({})`.

examples/ethereum_json_rpc/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+                )
+            ),
+            transform_method_name: 'eth_block_number_transform'
+        }).with_cycles(cycle_cost_total);
+
+    if (!ok(http_result)) {
+        ic.trap(http_result.err ?? 'http_result had an error');
+    }
+
+    return decodeUtf8(Uint8Array.from(http_result.ok.body));
```
dansteren, line 61: This encode/decode process is making me wonder if we should provide a middlewares concept. I've created https://github.com/demergent-labs/azle/issues/622 to explore ways to improve the DX here.

bdemann (dismissed):
LGTM

canisters/management.ts
```diff
@@ -124,6 +124,32 @@ export type ProvisionalTopUpCanisterArgs = {
...
+    headers: HttpHeader[];
+    body: Opt<blob>;
+    transform_method_name: Opt<string>;
+};
+
+export type HttpMethod = Variant<{
+    GET: null;
+    HEAD: null;
+    POST: null;
+}>;
```
dansteren, line 140: Consider exporting a const of the same name. i.e.

```ts
export const HttpMethod = {
    GET: { GET: null },
    HEAD: { HEAD: null },
    POST: { POST: null }
};
```
_(Not sure of on the casing here, they might just be `Get`, `Head`, `Post`, but something like this)_

Then when you're making a call you can do:

```ts
yield ManagementCanister.http_request({
    url: `https://xkcd.com/642/info.0.json`,
    http_method: HttpMethod.GET,
    ...
}).with_cycles(cycle_cost_total);
```

examples/outgoing_http_requests/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export function xkcd_transform(
+    http_response: HttpResponse
+): Query<HttpResponse> {
+    return {
+        ...http_response,
+        headers: []
+    };
+}
```
dansteren, line 82: Will this transform be common enough that we should just add it into azle? Like we could just let people import this as `ignore_headers` and then re-export it on their canister.

examples/outgoing_http_requests/test/test.ts
```diff
@@ -0,0 +1,68 @@
...
+                    result_json.num === expected_json.num &&
+                    result_json.link === expected_json.link &&
+                    result_json.year === expected_json.year &&
+                    result_json.news === expected_json.news &&
+                    result_json.safe_title === expected_json.safe_title &&
+                    result_json.alt === expected_json.alt &&
+                    result_json.img === expected_json.img &&
+                    result_json.title === expected_json.title &&
+                    result_json.day === expected_json.day
+            };
```
dansteren, line 63: These examples are basically duplicates. I would consider the following:

- Store the duplicated `expected_json` in a common constant variable
- Abstract the deep_comparison into it's own method
- Abstract the entire test into a common method to be shared between the two tests

Unless we are optimizing for speed right now.

dansteren (commented):
Left a lot of thoughts so I'm not approving or denying. Discussing how to proceed on a call may be easier.

examples/outgoing_http_requests/test/test.ts
lastmjs: I will fix this

lastmjs: This has been resolved

Issue #565, opened 2022-07-27
Implement Bitcoin functionality


lastmjs:
- [x] Provide importable types in Azle
    - [x] [`bitcoin_get_utxos`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_utxos)
    - [x] [`bitcoin_get_balance`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_balance)
    - [x] [`bitcoin_send_transactions`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_send_transaction)
    - [x] [`bitcoin_get_current_fee_percentiles`](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin_get_current_fee_percentiles)
- [ ] Simple example with tests
- [ ] npm packages (either find ones that work, fork, or implement our own)
- [ ] Documentation
    - [ ] Add jsdoc comments around types
- [ ] Complex example (if you want to)

Resources:
- https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-bitcoin-api

dansteren:
Quick progress update.

I've created all the necessary TS types in https://github.com/demergent-labs/azle/pull/627. Once that's merged in you should no longer need to declare your own types. You can just import the management canister and start calling methods on it directly.

Example canister:

```ts
import { CanisterResult, Update, Variant } from 'azle';
import {
    BitcoinNetwork,
    ManagementCanister,
    Satoshi
} from 'azle/canisters/management';

export type ExecuteGetBalanceResult = Variant<{
    ok: Satoshi;
    err: string;
}>;

export function* get_balance(address: string): Update<ExecuteGetBalanceResult> {
    const BITCOIN_API_CYCLE_COST = 100_000_000n;

    const canister_result: CanisterResult<Satoshi> =
        yield ManagementCanister.bitcoin_get_balance({
            address,
            min_confirmations: null,
            network: BitcoinNetwork.Regtest
        }).with_cycles(BITCOIN_API_CYCLE_COST);

    return canister_result;
}
```

I'm still working on creating a fully functioning example. I have one mostly working on the [565_bitcoin](https://github.com/demergent-labs/azle/tree/565_bitcoin) branch.

Where I'm currently hitting snags:
[...]

PR #627, opened 2022-08-12
Added bitcoin types to management canister

dansteren:
This adds types to the management canister. These types have been tested locally but because testing isn't straightforward, I'm proposing that we merge these in for right now and then add solid tests afterwards.

Some things complicating the testing include:
- `bitcoind` and `bitcoin-cli` must be installed locally
- `bitcoind` must be running at the time the dfx replica is started
- `bitcoin-cli` must be used to get the blockchain into a specific state before the tests are run

For these reasons I'm proposing we merge just these types in for now.

I've tested all of these APIs and verified that they in addition to not erring out, they also return data correctly.

Partial implementation of #565

### DIFF
--- a/canisters/management/bitcoin.ts
+++ b/canisters/management/bitcoin.ts
@@ -1,0 +1,71 @@
+import { blob, nat32, nat64, Opt, Variant } from '../../index';
+
+export type BitcoinAddress = string;
+export type BlockHash = blob;
+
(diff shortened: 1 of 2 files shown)


bdemann (approved):

dansteren:
These changes are incorporated in #631 and are merged in there.

PR #680, opened 2022-09-14
674 azle compiler refactor

lastmjs:


bdemann:
I merged this into #655 so that we would only have to fix conflicts once, so I am now closing this in favor of #655

PR #687, opened 2022-09-19
Organize canister methods

dansteren:
This separates the ACT node building from the token stream generation. It also incidentally does the same for the IC object.


src/compiler/typescript_to_rust/azle_generate/src/cdk_act/nodes/canister_method.rs
```diff
@@ -0,0 +1,138 @@
+use proc_macro2::{Ident, TokenStream};
+use quote::{format_ident, quote};
+
+use crate::cdk_act::{ActNode, ToTokenStream};
+
+#[derive(Clone)]
+pub enum CanisterMethodActNode {
```
lastmjs, line 7: I'm thinking that CanisterMethodActNode should be within the ActNode enum, just like any of the other types of an ActNode, since CanisterMethodActNode is an ActNode

lastmjs: Each ActNode variant should have other ACT nodes in them, like PrimitiveInfo I would rename to PrimitiveActNode (maybe @bdemann is taking care of this).

lastmjs (changes requested):

lastmjs (approved):

PR #700, opened 2022-09-26
686 create data type act nodes

bdemann:


bdemann:
This pr will never be merged. The big rename got moved to a different branch and the TypeAliasOrLiteral will be done later.
This would have to go through a lot of merge conflicts to get up-to-date so I think it will be easier to just use this pr as reference when implementing TypeAliasOrLiteral in the future

lastmjs:
Closed in favor of #705

PR #742, opened 2022-10-21
77 azle type errors

bdemann:


bdemann:
I opened this pr because it was the most straight forward way to test commits ee176f4 and 22a8cff which kept failing because of a corrupted cache. Those two commits show that they failed their tests while the merge commit 6edb8e0 tests are passing just fine.

PR #770, opened 2022-11-01
standardize canister method names

bdemann:
to be snake case instead of camel case


lastmjs (dismissed):

lastmjs:
Something is wrong here @bdemann

lastmjs:
Seems like it could be cache issues again, can you try to get these tests to pass?

PR #829, opened 2022-12-15
Adds the new timer API

dansteren:
This implements the new [timer API](https://docs.rs/ic-cdk/latest/ic_cdk/timer/index.html) so that the following is possible:

```ts
import { Duration, ic, TimerId, Update } from 'azle';

export function set_timers(delay: Duration): Update<[TimerId, TimerId, TimerId]> {
    const functionTimerId = ic.set_timer(delay, callback);
    const inlineTimerId = ic.set_timer(delay, () => {
        console.log('inline callback called');
    });
    const repeatingTimer = ic.set_timer_interval(delay, callback);

    return [functionTimerId, inlineTimerId, repeatingTimer];
}

export function clear_timer(timer_id: TimerId): Update<void> {
    ic.clear_timer(timer_id);
    console.log(`timer ${timer_id} cancelled`);
}

function callback(): void {
    console.log('callback called');
}
```

> **Note**
> Inline callbacks are supported and function callbacks are not required to be exported externally as canister methods.


README.md
```diff
@@ -2162,6 +2163,74 @@ export function reply_raw(): UpdateManual<RawReply> {
...
+}
+```
+
+#### set timer interval
+
+Examples:
+
+-   [timers](/examples/timers)
+
+```python
```
lastmjs, line 2216: This should be ts not python

README.md
```diff
@@ -2162,6 +2163,74 @@ export function reply_raw(): UpdateManual<RawReply> {
...
+
+Examples:
+
+-   [timers](/examples/timers)
+
+```ts
+import { Duration, ic, TimerId, Update } from 'azle';
+
+export function set_timers(delay: Duration): Update<[TimerId, TimerId]> {
+    const functionTimerId = ic.set_timer(delay, callback);
```
lastmjs, line 2196: Let's use snake case for everything

README.md
```diff
@@ -2162,6 +2163,74 @@ export function reply_raw(): UpdateManual<RawReply> {
...
+-   [clear timer](#clear-timer)
+-   [set timer](#set-timer)
+-   [set timer interval](#set-timer-interval)
+
+#### clear timer
+
+Examples:
+
+-   [timers](/examples/timers)
+
```
lastmjs, line 2177: Can you add a more information section like elsewhere in the documentation and link to the timers documentation on internetcomputer.org? If you could do this for all of the APIs that would be great

examples/timers/src/timers.ts
```diff
@@ -0,0 +1,61 @@
+import { Duration, ic, int8, Query, TimerId, Update } from 'azle';
+
+export type StatusReport = {
```
lastmjs, line 3: Is there a reason these types are exported? Let's probably not export them

src/compiler/typescript_to_rust/azle_generate/src/generators/header/mod.rs
```diff
@@ -1,4 +1,8 @@
+mod timers;
+
 pub fn generate_header_code() -> proc_macro2::TokenStream {
+    let timers_module = timers::generate_timers_module();
```
lastmjs, line 4: It's strange to me that the timers code is in the header. It seems architecturally like the timers code is just like any other code and has no reason to be in the header. It will be confusing to have to go to the header section to modify the timers code. I understand maybe you had to do this because of the module's scope? I would like to address this

src/compiler/typescript_to_rust/azle_generate/src/generators/header/timers.rs
```diff
@@ -0,0 +1,110 @@
+pub fn generate_timers_module() -> proc_macro2::TokenStream {
```
lastmjs, line 1: The timers module should not be in the header directory/file structure-wise, it should just be at the top level inside of generators. I see at as equal to other  generators and thus it will be confusing to have to go to the header for this.

src/compiler/typescript_to_rust/azle_generate/src/generators/ic_object/functions/clear_timer.rs
```diff
@@ -0,0 +1,17 @@
+pub fn generate_ic_object_function_clear_timer() -> proc_macro2::TokenStream {
+    quote::quote! {
+        fn _azle_ic_clear_timer(
+            _this: &boa_engine::JsValue,
+            _aargs: &[boa_engine::JsValue],
+            _context: &mut boa_engine::Context
+        ) -> boa_engine::JsResult<boa_engine::JsValue> {
+            let timer_id_js_value: boa_engine::JsValue = _aargs.get(0).unwrap().clone();
```
lastmjs, line 8: There's no need for this explicit type annotation, let's remove it.

src/compiler/typescript_to_rust/azle_generate/src/generators/vm_value_conversion/try_from_vm_value.rs
```diff
@@ -361,6 +368,12 @@ pub fn generate_try_from_vm_value_impls() -> proc_macro2::TokenStream {
             }
         }
 
+        impl CdkActTryFromVmValue<Vec<ic_cdk::timer::TimerId>, &mut boa_engine::Context> for boa_engine::JsValue {
```
lastmjs, line 371: Just remember that this will need to be changed once we merge in the list of lists code

src/compiler/typescript_to_rust/azle_generate/src/generators/vm_value_conversion/try_into_vm_value.rs
```diff
@@ -371,6 +378,12 @@ pub fn generate_try_into_vm_value_impls() -> proc_macro2::TokenStream {
             }
         }
 
+        impl CdkActTryIntoVmValue<&mut boa_engine::Context, boa_engine::JsValue> for Vec<ic_cdk::timer::TimerId> {
```
lastmjs, line 381: Just remember that this will need to be changed once we merge in the list of lists code

src/compiler/typescript_to_rust/azle_generate/src/generators/header/timers.rs
```diff
@@ -0,0 +1,110 @@
...
+            use ic_cdk::timer::TimerId;
+            use boa_engine::object::JsObject;
+            use core::time::Duration;
+
+            pub struct TimerCallback {
+                pub function: JsObject,
+                pub timer_id: TimerId,
+            }
+
+            thread_local! {
```
lastmjs, line 14: I think that all global variables should be stored in exactly one place, this is a best practice handed down by DFINITY and I think we should keep it

src/compiler/typescript_to_rust/azle_generate/src/generators/header/timers.rs
```diff
@@ -0,0 +1,110 @@
...
+                function: JsObject,
+                repeat: bool
+            ) -> TimerId {
+                let callback_id = uuid::Uuid::new_v4().to_string();
+
+                // We cannot pass the function directly to the closure because it's lifetime isn't
+                // long enough. It will go out of scope before it can be used by the closure. So
+                // instead, we store it in a global variable using a string key, and then later
+                // use that key (which can be passed to the closure) to look up the function.
+                TIMER_CALLBACKS_REF_CELL.with(|timer_callbacks_ref_cell| {
```
lastmjs, line 46: Why do we need to do this here?

src/compiler/typescript_to_rust/azle_generate/src/generators/header/timers.rs
```diff
@@ -0,0 +1,110 @@
...
+                move || {
+                    unsafe {
+                        let mut _azle_boa_context = crate::BOA_CONTEXT_OPTION.as_mut().unwrap();
+
+                        let timer_id = TIMER_CALLBACKS_REF_CELL.with(|timer_callbacks_ref_cell| {
+                            let timer_callbacks = timer_callbacks_ref_cell.borrow();
+
+                            let timer_callback = timer_callbacks.get(&callback_id).unwrap();
+
+                            crate::_azle_handle_boa_result(
```
lastmjs, line 90: We need to discuss what to do here,  because it might panic, and we could leave dangling state

lastmjs (changes requested):

lastmjs:
One more thing that we should address, let's test out JS closures and just make sure that state is captured appropriately. In case we forget, let's do this in Kybra as well.

index.ts
```diff
@@ -143,6 +148,29 @@ type ic = {
...
+     * To cancel the interval timer, pass the returned `TimerId` to `clear_timer`.
+     * Note that timers are not persisted across canister upgrades.
+     *
+     * @param interval The interval (in seconds) between each callback execution.
+     * @param callback the function to invoke after the specified delay has passed.
+     * @returns the ID of the created timer. Used to cancel the timer.
+     */
+    set_timer_interval: (
+        interval: Duration,
+        callback: () => Update<void>
```
lastmjs, line 172: Why is this wrapped in Update?

lastmjs (changes requested):

README.md
dansteren: The documentation on this hasn't been published by Dfinity yet. It's available at https://github.com/dfinity/interface-spec/blob/master/spec/index.adoc#global-timer, but let's wait to reference it till it's on https://internetcomputer.org/docs/current/references/ic-interface-spec.

dansteren:
> One more thing that we should address, let's test out JS closures and just make sure that state is captured appropriately. In case we forget, let's do this in Kybra as well.

:+1: I've updated the test to have a closure that captures a value like in Kybra.

lastmjs (approved):

PR #868, opened 2023-01-19
Improving local Rust installation

lastmjs:


install_rustup.sh
```diff
@@ -1,9 +1,10 @@
...
 # TODO If we want to save a little bit of time we might be able to instruct rustup to not install some components initially, like clippy and docs
+# TODO we might want to implement all of this in Node.js in the future for platform-independence etc
 
-azle_version=$(node -p "require('./package.json').version")
-rust_version=$(node -p "require('./package.json').rust_version")
+azle_version=$1
+rust_version=$2
 
 global_azle_config_dir=~/.config/azle/$azle_version
 global_azle_bin_dir=$global_azle_config_dir/bin
```
dansteren, line 10: `global_azle_target_dir=$global_azle_config_dir/target` is never actually used. Consider removing it.

dansteren, line 10: `global_azle_target_dir=$global_azle_config_dir/target` is never actually used. Consider removing it.

install_rustup.sh
```diff
@@ -15,28 +16,58 @@ global_azle_target_dir=$global_azle_config_dir/target
 export CARGO_HOME=$global_azle_config_dir
 export RUSTUP_HOME=$global_azle_config_dir
 
+function install_wasm32_unknown_unknown() {
+    $global_azle_rustup_bin target add wasm32-unknown-unknown &> $global_azle_logs_dir/step_2_wasm32_unknown_unknown
```
dansteren, line 19: The VS Code extension [ShellCheck](https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck) recommends double quoting when using variables to prevent globbing and word splitting.

I think that means:

```suggestion
    $global_azle_rustup_bin target add wasm32-unknown-unknown &> "$global_azle_logs_dir/step_2_wasm32_unknown_unknown"
```

This would apply anywhere you're using a variable I think.

install_rustup.sh
```diff
@@ -15,28 +16,58 @@ global_azle_target_dir=$global_azle_config_dir/target
...
+if ! test -e $global_azle_bin_dir/ic-wasm; then
+    echo -e "\nAzle $azle_version is installing ic-wasm\n"
+
+    install_ic_wasm
+fi
+
+if ! test -e $global_azle_bin_dir/ic-cdk-optimizer; then
+    echo -e "\nAzle $azle_version is installing ic-cdk-optimizer\n"
+
+    install_ic_cdk_optimizer
```
dansteren, line 71: I don't understand why there are two places where we install wasm32_unknown_unknown, ic_wasm and ic_cdk_optimizer. Is there a way to DRY this up?

src/azle.ts
```diff
@@ -34,23 +34,42 @@ const GLOBAL_AZLE_TARGET_DIR = `${GLOBAL_AZLE_CONFIG_DIR}/target`;
 azle();
 
 function azle() {
+    const install_rustup_path = path.resolve(__dirname, '../install_rustup.sh');
+
+    // TODO research exec vs spawn
+    execSync(`${install_rustup_path} ${version} ${rust_version}`, {
+        stdio: 'inherit'
+    });
```
dansteren, line 42: Running this script could fail. We should add error handling and tell them to check the logs directory.

src/azle.ts
```diff
@@ -34,23 +34,42 @@ const GLOBAL_AZLE_TARGET_DIR = `${GLOBAL_AZLE_CONFIG_DIR}/target`;
...
+
     const stdioType = getStdIoType();
     const canisterName = unwrap(getCanisterName(process.argv));
+    const canister_path = `.azle/${canisterName}`;
```
dansteren, line 46: This is stylistic so probably not worth changing, but you're mixing camel and snake case. I think this file is written all in camelCase up till now.

dansteren (changes requested):

PR #877, opened 2023-01-24
Void and null

bdemann:


examples/func_types/canisters/func_types/func_types.did
```diff
@@ -67,4 +67,7 @@ service : () -> {
     ) query;
```
lastmjs, line 67: Have we tested null as a param? If that doesn't work, then I think we should add a test case

test/index.ts
```diff
@@ -96,6 +96,34 @@ export async function run_tests(tests: Test[]) {
     }
 }
 
+export async function deploy_and_generate(
```
lastmjs, line 99: I think we should consider getting rid of this, it only works for single canisters, thus many examples will still need a different solution. This happened in the past, and it became simpler to just copy-paste across all examples

lastmjs (changes requested):

lastmjs:
There were many examples changed in Kybra, but I didn't see as many changed here

lastmjs (dismissed):

Issue #873, opened 2023-01-23
Cross canister calls in timers


lastmjs:
- [x] Ensure that uuids are actually random for promise implementation
- [x] Make sure promises won't walk over each other: Panicked at 'called `Option::unwrap()` on a `None` value', src/src/lib.rs:6663:59
- [x] Really seems like each promise needs its own uuid somehow
- [x] shouldn't each cross canister call promise get its own uuid? Why am I setting it at the beginning of the call?

PR #879, opened 2023-01-25
Timers cross canister

lastmjs:
Closes https://github.com/demergent-labs/azle/issues/873


src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
```diff
@@ -39,7 +39,8 @@ pub fn generate(canister_methods: &Vec<ActCanisterMethod>) -> TokenStream {
 
                         match _azle_method_name {
                             #(#match_arms)*
-                            _ => panic!("This cannot happen")
+                            "TIMER" => {},
```
dansteren, line 42: Could users ever define a "TIMER" method that would clash with this?

dansteren (approved):

bdemann (approved):

bdemann:
LGTM

dansteren:
New thought... do we need to update the typescript types for these methods now that they can take a promise? Currently we have:

```ts
    /**
     * Sets callback to be executed later, after delay. Panics if `delay` + time() is more than 2^64 - 1.
     * To cancel the timer before it executes, pass the returned `TimerId` to `clear_timer`.
     * Note that timers are not persisted across canister upgrades.
     *
     * @param delay The time (in seconds) to wait before executing the provided callback.
     * @param callback the function to invoke after the specified delay has passed.
     * @returns the ID of the created timer. Used to cancel the timer.
     */
    set_timer: (delay: Duration, callback: () => void) => TimerId;
    /**
     * Sets callback to be executed every interval. Panics if `interval` + time() is more than 2^64 - 1.
     * To cancel the interval timer, pass the returned `TimerId` to `clear_timer`.
     * Note that timers are not persisted across canister upgrades.
     *
     * @param interval The interval (in seconds) between each callback execution.
     * @param callback the function to invoke after the specified delay has passed.
     * @returns the ID of the created timer. Used to cancel the timer.
     */
    set_timer_interval: (interval: Duration, callback: () => void) => TimerId;
```
We may want to indicate that they can be async functions somehow.

PR #891, opened 2023-02-02
Changed canister method parsing to support new custom decorator annotations

dansteren:
This updates how we annotate canister methods from:

```ts
export function hello_world(): Query<string> {
    return "Hello World";
}
```
to

```ts
$query;
export function hello_world(): string {
    return "Hello World";
}
```


dansteren:
There are two commits that I think are worth double checking as part of the review for this:
- [Expose eprintln! statements in verbose mode](https://github.com/demergent-labs/azle/pull/891/commits/a8a5fd419b7d255acf950987171eeeda48b1b3dd)
- [Add compile-only mode to azle for rapid debugging](https://github.com/demergent-labs/azle/pull/891/commits/e76a48a8cab34dbf839c8fcebe1509111a8d563f)

These were both utility functionality that I added to help me as I was debugging. I thought they were useful but we can revert them if they don't fit.

lastmjs:
There shouldn't be this many file changes. It looks like this is because all of the dfx_generated folder have been put back. I explicitly removed all of the dfx_generated folders, so I'm not sure why they're back in here. They should be completely gone.

lastmjs:
It looks like implicit voids aren't being handled: `Canister methods must specify a return type`

When I have this code:

```typescript
import { $query, $update } from 'azle';

// This is a global variable that is stored on the heap
let message: string = '';

// Query calls complete quickly because they do not go through consensus
$query;
export function get_message(): string {
    return message;
}

// Update calls take a few seconds to complete
// This is because they persist state changes and go through consensus
$update;
export function set_message(new_message: string) {
    message = new_message; // This change will be persisted
}
```

I shouldn't have to set a type on set_message.

lastmjs:
We should go through the examples and get rid of all of the explicit voids, I don't think they should be there unless they're wrapped by an outer type like `Manual` or `Promise`. Our tests should have caught this unless you made the choice to only do explicit voids.

lastmjs (changes requested):
Would it make sense to reopon this against custom_decorators? Then I could see all of your changes specifically

PR #896, opened 2023-02-07
Fixed errors for mismatched return values

dansteren:
Previously when a function was declared as `null` we would just drop the returned value rather than trapping if the value wasn't `null`. Same for `void` functions that didn't return undefined. Now those will be errors.

Additionally, this changes the wording of uncaught exceptions from `Azle runtime error: {x}` to `Uncaught {x}`. This more closely aligns with Node and browsers.

Other minor changes include:
- switching to using a guard clause in src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
- renaming `_azle_handle_boa_result` to `_azle_unwrap_boa_result`
- renaming `_azle_handle_boa_error` to `_azle_js_value_to_string`

### DIFF
--- a/src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
+++ b/src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
@@ -15,61 +15,60 @@ pub fn generate(canister_methods: &Vec<ActCanisterMethod>) -> TokenStream {
             _azle_manual: bool
         ) -> boa_engine::JsValue {
             if
-                _azle_boa_return_value.is_object() == true &&
-                _azle_boa_return_value.as_object().unwrap().is_promise() == true
+                !_azle_boa_return_value.is_object() ||
+                !_azle_boa_return_value.as_object().unwrap().is_promise()
             {
-                // This runs all pending promises to completion
-                // TODO use the better Boa API once it's available
-                _azle_boa_context.eval("");
-
-                let object = _azle_boa_return_value.as_object().unwrap().borrow();
-                let promise = object.as_promise().unwrap();
-
-                return match &promise.promise_state {
-                    boa_engine::builtins::promise::PromiseState::Fulfilled(js_value) => {
-                        PROMISE_MAP_REF_CELL.with(|promise_map_ref_cell| {
-                            let mut promise_map = promise_map_ref_cell.borrow_mut();
-
-                            promise_map.remove(_azle_uuid);
-                        });
-
-                        if _azle_manual == true {
-                            return _azle_boa_return_value.clone();
-                        }
-
-                        match _azle_method_name {
-                            #(#match_arms)*
-                            "_AZLE_TIMER" => {},
-                            _ => panic!("method name was not found")
-                        };
-
-                        return _azle_boa_return_value.clone();
-                    },
-                    boa_engine::builtins::promise::PromiseState::Rejected(js_value) => {
-                        PROMISE_MAP_REF_CELL.with(|promise_map_ref_cell| {
-                            let mut promise_map = promise_map_ref_cell.borrow_mut();
-
-                            promise_map.remove(_azle_uuid);
-                        });
-
-                        let error_message = _azle_handle_boa_error(js_value.clone(), _azle_boa_context);
-
-                        panic!("Azle runtime error: {}", error_message);
-                    },
-                    boa_engine::builtins::promise::PromiseState::Pending => {
-                        PROMISE_MAP_REF_CELL.with(|promise_map_ref_cell| {
-                            let mut promise_map = promise_map_ref_cell.borrow_mut();
-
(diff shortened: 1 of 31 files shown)


src/compiler/typescript_to_rust/azle_generate/src/generators/boa_error_handlers.rs
```diff
@@ -3,21 +3,21 @@ use quote::quote;
...
             context: &mut boa_engine::Context
         ) -> boa_engine::JsValue {
             match boa_result {
                 Ok(_azle_boa_return_value) => _azle_boa_return_value,
                 Err(_azle_boa_error) => {
-                    let error_message = _azle_handle_boa_error(_azle_boa_error.to_opaque(context), context);
+                    let error_message = _azle_js_value_to_string(_azle_boa_error.to_opaque(context), context);
 
-                    panic!("Azle runtime error: {}", error_message);
+                    ic_cdk::api::trap(&format!("Uncaught {}", error_message));
```
lastmjs, line 15: Why did you make the decision to trap here?

src/compiler/typescript_to_rust/azle_generate/src/generators/boa_error_handlers.rs
```diff
@@ -3,21 +3,21 @@ use quote::quote;
...
                 },
             }
         }
 
-        fn _azle_handle_boa_error(error_value: boa_engine::JsValue, context: &mut boa_engine::Context) -> String {
+        fn _azle_js_value_to_string(error_value: boa_engine::JsValue, context: &mut boa_engine::Context) -> String {
```
lastmjs, line 20: I don't think this is an appropriate name, it sounds like it's a generic JsValue -> String function. It's specifically turning a JS error into a string

src/compiler/typescript_to_rust/azle_generate/src/generators/canister_methods/query_and_update.rs
```diff
@@ -67,25 +67,28 @@ fn generate_return_expression(fn_decl: &AzleFnDecl) -> proc_macro2::TokenStream
...
-fn type_is_null_or_void(ts_type: &TsType) -> bool {
-    match ts_type {
+    let null_and_void_handler = match return_type {
         TsType::TsKeywordType(keyword) => match keyword.kind {
-            // TODO: Consider handling `TsNeverKeyword` and `TsUndefinedKeyword`
-            TsNullKeyword | TsVoidKeyword => true,
-            _ => false,
+            TsNullKeyword => quote! {
+                if !_azle_final_return_value.is_null() {
+                    ic_cdk::api::trap("TypeError: value is not of type 'null'");
```
lastmjs, line 74: Why the decision to trap everywhere?

lastmjs, line 74: Are you sure that TypeError is the appropriate error for this situation?

src/compiler/typescript_to_rust/azle_generate/src/generators/canister_methods/query_and_update.rs
```diff
@@ -67,25 +67,28 @@ fn generate_return_expression(fn_decl: &AzleFnDecl) -> proc_macro2::TokenStream
...
+                }
+            },
+            TsVoidKeyword => quote! {
+                if !_azle_final_return_value.is_undefined() {
+                    ic_cdk::api::trap("TypeError: value is not of type 'void'");
```
lastmjs, line 79: I think we might want to change this error to describe that undefined must be returned explicitly or implicitly

lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
```diff
@@ -15,61 +15,60 @@ pub fn generate(canister_methods: &Vec<ActCanisterMethod>) -> TokenStream {
             _azle_manual: bool
         ) -> boa_engine::JsValue {
             if
-                _azle_boa_return_value.is_object() == true &&
-                _azle_boa_return_value.as_object().unwrap().is_promise() == true
+                !_azle_boa_return_value.is_object() ||
```
lastmjs, line 18: Is this the only material change here?

lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/generators/boa_error_handlers.rs
dansteren: Better DX

src/compiler/typescript_to_rust/azle_generate/src/generators/async_await_result_handler.rs
dansteren: yes. Just changing to a guard clause.

lastmjs (approved):

PR #898, opened 2023-02-16
Changed external canister parsing

dansteren:
External canisters are now declared like this:

```ts
import { CanisterResult, ExternalCanister, query } from 'azle';

export class Notifier extends ExternalCanister {
    @query
    get_notifier: () => CanisterResult<string>;
}
```

And called like this:
```ts
const notifiers_canister: Notifier = new Notifier(
    Principal.fromText('ryjl3-tyaaa-aaaaa-aaaba-cai')
);
const result = await notifiers_canister.get_notifier().call();
```


index.ts
```diff
@@ -230,10 +226,36 @@ export type Stable64GrowResult = Variant<{
...
+/**
+ * A decorator for annotating query methods on external canisters. Can only be
+ * used on class properties with a return type of (args: any[]) =>
+ * CanisterResult<T>.
+ *
+ * @example
+ * ```ts
+ * export class SomeOtherCanister extends ExternalCanister {
+ *   @query
+ *   someCanisterMethod: (someParam: SomeParamType) => CanisterResult<SomeReturnType>;
```
lastmjs, line 238: Method name should be camel case

index.ts
```diff
@@ -230,10 +226,36 @@ export type Stable64GrowResult = Variant<{
     err: StableMemoryError;
 }>;
 
+/**
+ * A decorator for annotating query methods on external canisters. Can only be
```
lastmjs, line 230: A decorator should probably `decorate` and not `annotate`. It's especially important to make the distinction because we're calling our custom `$query` and `$update` things annotations

index.ts
```diff
@@ -289,6 +311,27 @@ function external_canister_method_decoration(target: any, name: string) {
...
 }
 
+/**
+ * Parent class for creating Canister definitions. To create an external
+ * canister extend this class.
+ * @example
+ * ```ts
+ * export class SomeOtherCanister extends ExternalCanister {
+ *   @query
+ *   someCanisterMethod: (someParam: SomeParamType) => CanisterResult<SomeReturnType>;
```
lastmjs, line 321: Method should be camel case

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/source_map/private_get_source_file_info.rs
```diff
@@ -0,0 +1,37 @@
+use swc_common::{Loc, SourceMap, Span};
+
+use crate::ts_ast::source_map::get_source_file_info::GetSourceFileInfo;
+
+pub(super) trait PrivateGetSourceFileInfo {
```
lastmjs, line 5: Let's just not go here, I've only seen this type of visibility cause pain (in my limited experience)

lastmjs, line 5: Why is it named thusly? Let's discuss this

lastmjs (changes requested):
All of the tests are failing, I think I know why. I'm curious to know how you tested it locally, because I don't see how any tests would pass locally or in the GitHub Action

index.ts
dansteren: meaning snake case... yes, I can change that. Will change.

dansteren:
I'll go back and make sure that any invalid properties on an external canister cause an error rather than just silently not being parsed.

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/source_map/private_get_source_file_info.rs
dansteren: I will remove super.

dansteren: postponing addressing this... :grin:

lastmjs (approved):

PR #901, opened 2023-02-18
Changed `Record` syntax

dansteren:
Records are now expected wrapped in a `Record` type, instead of just an object literal.

Example:

```ts
type SomeRecord = Record<{
  some_prop: string
}>
```


examples/inline_types/src/types.ts
```diff
@@ -10,17 +10,18 @@ import {
...
     query,
+    Record,
     update,
     Variant
 } from 'azle';
 
-export type User1 = {
+export type User1 = Record<{
     id: string;
     job: {
```
lastmjs, line 20: We should discuss the inline records having the `Record` wrapping as well, just like inline variants.

lastmjs (changes requested):

examples/inline_types/src/types.ts
```diff
@@ -10,17 +10,18 @@ import {
     nat,
     nat64,
     query,
+    Record,
```
lastmjs, line 13: The parsing should break if the type is not wrapped in `Record`.

lastmjs (changes requested):

examples/inline_types/src/types.ts
dansteren: I've updated the parser to handle/require this.

examples/inline_types/src/types.ts
dansteren: I've updated the parser to handle/require this.

canisters/ledger/index.ts
```diff
@@ -221,39 +252,28 @@ export type QueryBlocksResponse = {
...
-        // The number of blocks that can be fetch using the callback.
-        length: nat64;
-
-        // The function that should be called to fetch the archived blocks.
-        // The range of the blocks accessible using this function is given by [from]
-        // and [len] fields above.
-        callback: QueryArchiveFn;
-    }[];
-};
+    archived_blocks: ArchivedBlock[];
```
lastmjs, line 255: Let's revert these to inline types

lastmjs (changes requested):

lastmjs (approved):

canisters/ledger/index.ts
dansteren: Done

PR #946, opened 2023-03-16
Add nested tuple example

dansteren:
Also adds example to ensure that complex types work even if only referenced in system canister methods, and not also in query/update methods. This used to be a bug but it looks to be working! :raised_hands:

### DIFF
--- a/examples/tuple_types/test/tests.ts
+++ b/examples/tuple_types/test/tests.ts
@@ -495,6 +495,26 @@ export function get_tests(
                     ok: result[0].hello === 0n && result[1].goodbye === 1n
                 };
             }
+        },
+        {
+            name: 'nested tuple test',
+            test: async () => {
+                let expected_result: [[string, [number, number]], bigint] = [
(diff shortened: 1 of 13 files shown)


examples/tuple_types/test/tests.ts
```diff
@@ -495,6 +495,26 @@ export function get_tests(
                     ok: result[0].hello === 0n && result[1].goodbye === 1n
                 };
             }
+        },
+        {
+            name: 'nested tuple test',
+            test: async () => {
+                let expected_result: [[string, [number, number]], bigint] = [
```
lastmjs, line 502: Is there a reason this is let instead of const?

lastmjs (changes requested):

examples/tuple_types/test/tests.ts
dansteren: nope. Good catch :)

dansteren: nope. good catch!

PR #961, opened 2023-03-21
Added candid syntax improvements

dansteren:
Adds:
* A new `text` datatype as an alias for `string`. Both can be used to represent the candid `text` type.
* `number` as an alias to the candid `float64` type.
* Enforcement of return types on all canister methods, and `void` for system methods (and `Promise<void>` for heartbeat)


lastmjs:
I'm considered about promise void

examples/func_types/canisters/func_types/func_types.ts
```diff
@@ -46,7 +46,7 @@ type NullFunc = Func<
 >;
 
 $init;
-export function init_() {
+export function init_(): void {
```
lastmjs, line 49: Rename this to `init`.

the_azle_book/src/reference/canister_methods/heartbeat.md
```diff
@@ -11,7 +11,7 @@ Examples:
 import { $heartbeat } from 'azle';
```
lastmjs, line 11: Run The Azle Book build command

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/azle_type/azle_type_ref/to_data_type.rs
```diff
@@ -29,9 +29,11 @@ impl AzleTypeRef<'_> {
             "nat16" => CandidType::Primitive(Primitive::Nat16),
             "nat32" => CandidType::Primitive(Primitive::Nat32),
             "nat64" => CandidType::Primitive(Primitive::Nat64),
+            "number" => CandidType::Primitive(Primitive::Float64),
```
lastmjs, line 32: I think this should/can be removed

lastmjs (changes requested):

lastmjs (approved):

PR #971, opened 2023-03-31
unify guard functions, imports and manual reply

bdemann:


examples/guard_functions/src/index.ts
dansteren, line 1: Let's add this back in. The guard property should be a string though and the only way to add it back in is to disable prettier when committing. You can do that with the `--no-verify` flag.

We should either:
- Add this as either an additional test in Azle (not in the normal set of tests)
- Filter this out in kybra.

> **Note**
> There might be a way to disable this in prettier with a nice comment. Like `//prettier:disable-oneline`... or something.

examples/guard_functions/test/tests.ts
```diff
@@ -156,79 +147,7 @@ export function getTests(
                     // Why it only says "Error" not "CustomError" I don't understand.
                     return {
                         Ok: (error as AgentError).message.includes(
-                            `Uncaught Error: Execution halted by \\"throwCustomError\\" guard function`
-                        )
-                    };
-                }
-            }
-        },
-        {
```
dansteren, line 165: We need to have these tests back in Azle. We can filter them out in Kybra.

dansteren (changes requested):
Just some minor things :)

examples/guard_functions/src/index.ts
dansteren: Looks like the syntax is `// prettier-ignore` before the line.

See: https://prettier.io/docs/en/ignore.html#javascript

PS: This was a great suggestion. I didn't even think about having prettier ignore this but this will be much better if we can get it to work. Thanks!

dansteren (approved):

PR #977, opened 2023-04-04
Generics for Variants, Result type

lastmjs:


src/compiler/typescript_to_rust/azle_generate/src/ts_ast/azle_type/azle_type_ref/to_data_type.rs
dansteren, line 1: This file is still called `to_data_type.rs`. We should maybe rename it to `to_candid_type` while we're at it.

src/compiler/typescript_to_rust/azle_generate/src/candid_type/variant/mod.rs
```diff
@@ -20,7 +24,39 @@ impl SourceMapped<'_, TsTypeAliasDecl> {
     pub fn to_variant(&self) -> Option<Variant> {
         self.process_ts_type_ref("Variant", |azle_type_ref| {
             let mut variant = azle_type_ref.to_variant();
-            variant.name = Some(self.id.get_name().to_string());
+
+            let name_string = self.id.get_name().to_string();
+
+            variant.name = Some(if name_string == "Result" { "_AzleResult".to_string() } else { name_string } );
+
+            let type_params = if let Some(type_params) = &self.type_params {
```
dansteren, line 32: I would break this out into a separate function...

src/compiler/typescript_to_rust/azle_generate/src/candid_type/variant/mod.rs
```diff
@@ -20,7 +24,39 @@ impl SourceMapped<'_, TsTypeAliasDecl> {
...
+                type_params
+                    .params
+                    .iter()
+                    .map(|type_param| TypeParam {
+                        name: type_param.name.get_name().to_string(),
+                        try_into_vm_value_trait_bound: quote!(
```
dansteren, line 38: These try_into and try_from trait bounds don't seem to belong at this level to me. What I'm thinking would likely be a significant effort so I think this is still probably the quickest way to do this.

But I'm thinking that the try_into bound will always be the same for Azle and so we could maybe set that once somewhere else, rather than on each param. And the the try_from, it seems like maybe the `Box<#name>` part is the only thing that's specific to each param. So yeah, just seems like this belongs somewhere else...

But again, this is likely the best way for now since doing something else would take some time. Just wanted to acknowledge this.

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/azle_type/azle_type_ref/to_data_type.rs
```diff
@@ -35,8 +35,29 @@ impl AzleTypeRef<'_> {
     }
 
     fn to_type_ref(&self) -> TypeRef {
+        let type_arguments = if let Some(type_params) = &self.ts_type_ref.type_params {
+            type_params
+                .params
+                .iter()
+                .map(|param| {
+                    // TODO is this the right source map to pass in?
```
dansteren, line 43: Yes, the source map is good. The source map is tied to a program so the same one should be used for a type_ref as for its type params.

PR #985, opened 2023-04-07
Reduced name collisions

dansteren:
Reduced collisions with user-defined names by
- moving most of our code into modules where it can't conflict with user-defined code
- removing as many `use` statements as possible

Modules:
- javascript
- ref_cells
- randomness
- vm_value_conversion
- ic_object
- stable_b_tree_map

Note: now that everything is in it's own module, we can use `use` statements, just not at the global level. Also, methods like those on the ic_object no longer need to be prefixed.


dansteren:
This undertaking was abandoned for now in favor of prefixing everything. I hit a wall with undrestanding when to prefix things with `crate::` and when not too. Particularly, there are a few types in the service_wrapper functions and in ic.reply that need to have the datatype that they are convert vm values into. Those values are sometimes builtins (e.g. i8, String, `()`), sometimes from candid (e.g. candid::Reserved), and sometimes user-defined and should be prefixed with crate `crate::SomeUserDefinedRecord`.

This is further complicated by the fact that type parameters may also fall into those same categories. So sometimes we would generate `Ok(crate::Ok)` when really it should have been the global `Ok`. 

Figuring out what each value is, where it comes from, and whether it needs to be module-namespace qualified get really complicated and we couldn't figure out a quick easy way to do this. Especially because sometimes a type does need to be qualified when it shows up in one module, but not when present in another.

For all these reasons we've decided to abandon using modules for now. Instead we will namespace qualify everything, rename all imported traits, prefix user-defined variables, and not have `use` statements.

PR #995, opened 2023-04-11
Fixed name collisions

dansteren:
In order to reduce conlicts with user-defined code, this PR:

- Removes global `use` statements in favor of fully qualified names
- Removes `_azle_` prefixes from variables in favor of prefixing user-defined variables with `cdk_user_defined_`
- Adds `_azle_` prefixes to global functions so they don't conflict with user-defined functions
- Removes leading underscores on function params where the param is used in the function body

> **Note**
> We still have a `use candid::{Encode, Decode}` statement, but it doesn't clash with structs/enums/types so users can still have their own struct/enum/type with that name.

Depends on https://github.com/demergent-labs/cdk_framework/pull/65

### DIFF
--- a/src/compiler/typescript_to_rust/azle_generate/src/candid_type/func/rust.rs
+++ b/src/compiler/typescript_to_rust/azle_generate/src/candid_type/func/rust.rs
@@ -18,7 +18,7 @@ pub fn generate_list_into_vm_value_impl(function_name: String) -> TokenStream {
     quote! {
         impl CdkActTryIntoVmValue<&mut boa_engine::Context, boa_engine::JsValue> for Vec<#function_name> {
             fn try_into_vm_value(self, context: &mut boa_engine::Context) -> Result<boa_engine::JsValue, CdkActTryIntoVmValueError> {
-                try_into_vm_value_generic_array(self, context)
+                _azle_try_into_vm_value_generic_array(self, context)
             }
         }
     }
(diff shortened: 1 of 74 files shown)


bdemann (approved):

src/compiler/typescript_to_rust/azle_generate/src/candid_type/func/rust.rs
```diff
@@ -18,7 +18,7 @@ pub fn generate_list_into_vm_value_impl(function_name: String) -> TokenStream {
     quote! {
         impl CdkActTryIntoVmValue<&mut boa_engine::Context, boa_engine::JsValue> for Vec<#function_name> {
             fn try_into_vm_value(self, context: &mut boa_engine::Context) -> Result<boa_engine::JsValue, CdkActTryIntoVmValueError> {
-                try_into_vm_value_generic_array(self, context)
+                _azle_try_into_vm_value_generic_array(self, context)
```
lastmjs, line 21: We shouldn't prefix our method names, instead we will prefix their methods and export them with the correct name using ic_cdk macros

lastmjs (approved):

src/compiler/typescript_to_rust/azle_generate/src/candid_type/func/rust.rs
dansteren: This will be done later. See https://github.com/demergent-labs/cdk_framework/issues/66

PR #1008, opened 2023-04-21
Changed Opt to be an algebraic data type

dansteren:
The syntax for Opt types hasn't changed. It's still `Opt<SomeTime>`. Creating opts has changed.

Opt values are now created using the Opt.Some and Opt.None helpers.

```ts
import {Opt} from "azle"

const someValue: Opt<string> = Opt.Some("some value");
const noneValue: Opt<string> = Opt.None
```

> **Note**
> The type annotations on `someValue` and `noneValue` aren't necessary, but rather for illustrative purposes

### DIFF
--- a/examples/cross_canister_calls/canisters/canister2/canister2.ts
+++ b/examples/cross_canister_calls/canisters/canister2/canister2.ts
@@ -50,7 +50,8 @@ export function balance(id: string): nat64 {
 
 $query;
 export function account(accountArgs: AccountArgs): Opt<Account> {
-    return state.accounts[accountArgs.id] ?? null;
+    let account = state.accounts[accountArgs.id];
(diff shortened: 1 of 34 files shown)


examples/cross_canister_calls/canisters/canister2/canister2.ts
```diff
@@ -50,7 +50,8 @@ export function balance(id: string): nat64 {
 
 $query;
 export function account(accountArgs: AccountArgs): Opt<Account> {
-    return state.accounts[accountArgs.id] ?? null;
+    let account = state.accounts[accountArgs.id];
```
lastmjs, line 53: const

lastmjs (dismissed):

Issue #223, opened 2022-04-13
Allow Azle to have the equivalent of process.env


lastmjs:
We can allow for environment variables probably by grabbing them during build time and creating the global process object.

lastmjs:
Perhaps we can start reading from dfx.json to get environment variables...this would not be standard, but could be awesome

PR #1009, opened 2023-04-25
Remove external optimization

lastmjs:
This improves the initial Azle prerequisites installation by 5-7 minutes, removes ic-cdk-optimizer and ic-wasm, solves the cmake and cc issues, gets rid of rustfmt command line stuff, cuts the step 1 time in half on subsequent builds

Closes: https://github.com/demergent-labs/azle/issues/1012 and https://github.com/demergent-labs/azle/issues/223


src/compiler/typescript_to_rust/azle_generate/src/main.rs
```diff
@@ -14,11 +16,31 @@ fn main() {
     let compiler_info_string = std::fs::read_to_string(compiler_info_path).unwrap();
     let compiler_info: CompilerInfo = serde_json::from_str(&compiler_info_string).unwrap();
 
+    let environment_variables = if let Some(envs) = args.get(2) {
```
lastmjs, line 19: Let's leave a comment here, consider reading the dfx.json file, remove as many command line arguments as possible

src/compiler/typescript_to_rust/azle_generate/src/main.rs
```diff
@@ -14,11 +16,31 @@ fn main() {
...
+        main_js,
+        &compiler_info.plugins,
+        &environment_variables,
+    )
+    .to_string();
 
-    // TODO let's fix this now too
-    // TODO let's really do this
-    println!("{}", result);
+    let mut f = File::create("../src/lib.rs").expect("Unable to create file");
```
lastmjs, line 43: Can we use rustfmt as a library right here? Then we could get rid of the rustfmt command line thing in the Node.js code

src/compiler/typescript_to_rust/generate_rust_canister.ts
```diff
@@ -24,61 +28,64 @@ export function generateRustCanister(
...
+            ...process.env,
+            CARGO_TARGET_DIR: GLOBAL_AZLE_TARGET_DIR,
+            CARGO_HOME: GLOBAL_AZLE_RUST_DIR,
+            RUSTUP_HOME: GLOBAL_AZLE_RUST_DIR
+        }
+    });
+
+    // TODO should we handle presenting errors to the user somehow here?
+    // TODO maybe we should tell them to just rerun with verbose if the status code is not 0
+    if (buildResult.status !== null && buildResult.status !== 0) {
```
lastmjs, line 74: Make sure to handle error and status code...handle any possible failure within reason

src/compiler/typescript_to_rust/generate_rust_canister.ts
```diff
@@ -24,61 +28,64 @@ export function generateRustCanister(
...
 
 function runAzleGenerate(
     compilerInfoPath: string,
     canisterPath: string,
-    rootPath: string
-): Result<Rust, AzleError> {
-    if (isVerboseMode()) {
-        console.info('running azle_generate');
+    canisterConfig: JSCanisterConfig
+) {
```
lastmjs, line 51: Put the result back in

src/compiler/utils/types.ts
```diff
@@ -19,8 +19,12 @@ export type JSCanisterConfig = Readonly<{
     ts: string;
     candid: string;
     wasm: string;
+    env: string[];
```
lastmjs, line 22: Should also be optional

src/compiler/typescript_to_rust/azle_generate/src/lib.rs
```diff
@@ -22,8 +22,9 @@ pub fn generate_canister(
     ts_file_names: &Vec<String>,
     main_js: String,
     plugins: &Vec<Plugin>,
+    environment_variables: &Vec<(String, String)>,
 ) -> TokenStream {
-    TsAst::new(ts_file_names, main_js, plugins)
+    TsAst::new(ts_file_names, main_js, plugins, environment_variables)
```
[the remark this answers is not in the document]
lastmjs: Resolved

src/compiler/typescript_to_rust/azle_generate/src/main.rs
lastmjs: Done, using a library instead

dansteren (approved):
It overall looks good. There are some polishes that could be done, different places/ways to handle errors and such but I'll tackle them in https://github.com/demergent-labs/azle/issues/1017

src/compiler/typescript_to_rust/generate_rust_canister.ts
```diff
@@ -6,192 +6,96 @@ import {
...
     const compilerInfoPath = join(
         canisterPath,
-        rootPath,
+        canisterConfig.root,
         'azle_generate',
         'compiler_info.json'
     );
 
+    // TODO why not just write the dfx.json file here as well?
     writeFileSync(compilerInfoPath, JSON.stringify(compilerInfo));
```
dansteren, line 34: Just had a thought. Just a mental viewpoint, not that we need to change the code. But it seems like we're writing a config file for azle_generate here. So maybe we call it `.azlegeneraterc`? Kinda makes sense to me. Then when you run azle_generate it always looks for that file so we don't have to pass a file name across the CLI.

This makes more sense to me than calling it `compiler_info.json` and having to specify it via the CLI each time.

Just something to think about.

src/compiler/install_rust_dependencies.ts
```diff
@@ -1,16 +1,57 @@
...
+    if (isWasm32TargetInstalled()) {
+        return;
+    }
 
-    execSync(`"${installRustDependenciesPath}" ${azleVersion} ${rustVersion}`, {
-        stdio: 'inherit'
+    time('[0/2] 🏗️  Preparing prerequisites...', 'inline', () => {
+        const installRustDependenciesPath = resolve(
+            __dirname,
+            '../../install_rust_dependencies.sh'
```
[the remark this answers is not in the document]
lastmjs: I've gone back and using join in many places that I wasn't, but sure please keep cleaning these up, let's just keep that in mind moving forward which I have been doing.

dansteren (commented):
Just wanted to surface these thoughts I has while reviewing. Just ideas for later.

PR #1019, opened 2023-05-04
Optimize azle generate

lastmjs:


src/compiler/typescript_to_rust/generate_rust_canister.ts
```diff
@@ -47,13 +49,57 @@ function runAzleGenerate(
...
+
+    return runBinary(
+        azleGenerateBinPath,
+        azleGenerateDir,
+        compilerInfoPath,
+        canisterConfig
+    );
+}
+
+function buildBinary(
```
dansteren, line 92: Although this works, it's not intuitive to me. This method builds the binary, and copies it to the correct location. Maybe we should separate those two pieces out? or rename for clarity?

It's also interesting to me that  on line 97 we create the parent directories just from a bunch of parts of the path, but the azleGenerateBinPath theoretically could be somewhere else. We should somehow tie those two together.

Something like:
mkdirSync(path.dirname(azleGenerateBinPath)).

Then I would also just change the names on all this stuff on the inside of this function. So my sig would be:

```ts
function buildBinary(cmdDir: string, src: string, dest: string): Result<null, SpawnSyncError>
```

Like, we just need to know we're copying from src to dest, not what those paths are.

I just have a hard time keeping track of which path is which when all their names are all so similar.

dansteren (approved):

src/compiler/typescript_to_rust/generate_rust_canister.ts
lastmjs: Great thoughts for the future

PR #1072, opened 2023-06-14
Added JS Docs for Candid Types and Canister Method Annotations

dansteren:
Improves the developer experience by providing relevant documentation at the hover of a mouse.

### DIFF
--- a/src/lib/candid_types/func.ts
+++ b/src/lib/candid_types/func.ts
@@ -1,8 +1,32 @@
-import { Principal } from './';
+import { Principal, Service } from './';
 
-export type Func<T> = T;
+/**
+ * Represents a reference to a {@link Service} method. The type param specifies
+ * the functional signature of the method but under the hood it is really just
+ * an array with two values:
+ *
+ * - the principal of the {@link Service} where the method is
+ * defined
+ * - the name of the method.
+ *
+ * See https://internetcomputer.org/docs/current/references/candid-ref#type-func---
+ *
+ * @typeParam T - the signature of the referenced {@link Service} method. Must
+ * be one of:
+ *
+ * - {@link Query}
+ * - {@link Update}
+ * - {@link Oneway}
+ */
+export type Func<T extends Mode> = T;
 
-// TODO type these more strictly
-export type Query<T extends (...args: any[]) => any> = [Principal, string];
-export type Update<T extends (...args: any[]) => any> = [Principal, string];
-export type Oneway<T extends (...args: any[]) => any> = [Principal, string];
+export type ServiceMethod = (...args: any[]) => any;
+
+export type Query<T extends ServiceMethod> = [Principal, string];
+export type Update<T extends ServiceMethod> = [Principal, string];
+export type Oneway<T extends ServiceMethod> = [Principal, string];
+
+export type Mode =
+    | Query<ServiceMethod>
+    | Update<ServiceMethod>
+    | Oneway<ServiceMethod>;
--- a/src/lib/candid_types/func.ts
+++ b/src/lib/candid_types/func.ts
@@ -11,22 +11,49 @@ import { Principal, Service } from './';
  *
  * See https://internetcomputer.org/docs/current/references/candid-ref#type-func---
  *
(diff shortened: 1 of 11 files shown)


src/lib/candid_types/func.ts
```diff
@@ -1,8 +1,59 @@
-import { Principal } from './';
+import { Principal, Service } from './';
 
-export type Func<T> = T;
+/**
+ * Represents a reference to a {@link Service} method. The type param specifies
+ * the functional signature of the method but under the hood it is really just
+ * an array with two values:
```
lastmjs, line 6: I think we should explain that it's a TypeScript tuple, it's not really an array. Well the type is a TypeScript tuple, at runtime it will be an array. I think mentioning the true TypeScript type would be beneficial.

src/lib/candid_types/index.ts
```diff
@@ -5,43 +5,186 @@ export { Func } from './func';
 export { Service } from './service';
 export { Variant } from './variant';
 
+/** Marks a type as being an alias to a candid type */
 export type Alias<T> = T;
+
+/**
+ * Represents a sequence of bytes. Synonymous with `Vec<nat8>`.
```
lastmjs, line 12: We need to check on `Vec<nat8>`, I'm not sure that will work. Even if it works technically, the TypeScript types would be incorrect. I don't think it's a best practice to share with people, and we should probably discourage and thus not mention it.

src/lib/candid_types/variant.ts
```diff
@@ -26,6 +26,25 @@ export type RequireExactlyOne<
...
     Omit<ObjectType, KeysType>;
 
+/**
+ * Represent a value that is from exactly one of the given cases or tags. If a
+ * tag does not carry data it is idiomatic to use the `null` type.
+ *
+ * @example
+ * ```ts
+ * type Shape = Variant<{
+ *   Dot: null,
```
lastmjs, line 36: These should end with semicolons I believe, not commas

src/lib/candid_types/index.ts
```diff
@@ -5,43 +5,186 @@ export { Func } from './func';
...
 };
 
 /**
- * Used to mark an object as a Candid record.
+ * An unordered heterogeneous collection of labeled values.
+ *
+ * @example
+ * ```ts
+ * type Address = Record<{
+ *   street: string,
```
lastmjs, line 149: These should end with semicolons I believe, not commas

src/lib/results.ts
```diff
@@ -1,32 +1,106 @@
...
+ *
+ * Each service method call can await a result ({@link CallResult.call}) or can
+ * be sent without awaiting a response ({@link CallResult.notify}).
+ *
+ * To send cycles with your request first chain either {@link CallResult.cycles}
+ * or {@link CallResult.cycles128}.
+ */
 export type CallResult<T> = {
+    /**
+     * Dispatches the asynchronous {@link Service} method call and awaits a
```
lastmjs, line 48: It doesn't await a response (depending on how you think about it, I think it might be confusing to say this), you must await the response. I think we should just mention that this returns a promise that can be awaited

src/lib/results.ts
```diff
@@ -1,32 +1,106 @@
...
+     *
+     * @param cycles - the number of cycles to send with the call
+     * @returns the {@link CallResult} instance with the provided cycle amount
+     * stored
+     */
     cycles128: (cycles: nat) => CallResult<T>;
 };
 
+/**
+ * A result containing the return value of the {@link Service} method call or
```
lastmjs, line 89: I'm not sure we can say that it's functionally equivalent, since this doesn't have the Ok and Err helpers. Also...why is this even exported? This is supposed to be sort of a hidden type, I wonder if we shouldn't export it or document it in this way, or if we do document it explain how it's an internal helper type.

lastmjs (changes requested):

src/lib/results.ts
dansteren: Fine. I'll just delete FinalCallResult so there.

PR #1083, opened 2023-06-21
Code Coverage: Added tests for aliases and other complex types

dansteren:
This updates the rust_type_conversions example and adds tests to ensure the syntax stays supported.


examples/rust_type_conversions/src/index.ts
```diff
@@ -237,60 +240,64 @@ type VariantWithInline = Variant<{
...
-                one: boolean;
-                two: boolean;
-                three: boolean;
-            }>,
-            inline_variant: Variant<{ one: string; two: null; three: boolean }>,
-            inline_func: Func<Query<(param1: string) => string>>
-        ) => string
-    >
->;
+// TODO: Funcs with inline types don't currently work
```
lastmjs, line 243: Do we have an issue for this?

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/ts_type/to_candid_type.rs
```diff
@@ -21,6 +21,15 @@ impl SourceMapped<'_, TsType> {
             TsType::TsTupleType(_) => {
                 return Err(vec![UnexpectedTsTupleTypes::from_ts_type(self).into()]);
             }
+            TsType::TsArrayType(_) => {
+                // TODO: Because this isn't handled types like this currently break:
```
lastmjs, line 25: I don't think this needs to be fixed right? You have to use `Vec`

lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/ts_ast/ts_type/to_candid_type.rs
dansteren: Correct. I've deleted this.

examples/rust_type_conversions/src/index.ts
dansteren: Just made one https://github.com/demergent-labs/azle/issues/1087

PR #1097, opened 2023-06-30
Changed runtime errors to bubble up correctly

dansteren:
Rather than panic any time we encounter an error at runtime, we are embracing results and exceptions. Errors at the topmost level will call `ic.trap` to surface their error messages. JavaScript exceptions will bubble up to that layer. Native rust functions called from JavaScript (like methods on the global `ic` object) will raise exceptions to the JavaScript layer above them.

> **Note**
> In the future, `TryIntoVmValue` should really return a `boa_engine::JsResult`. This was not attempted as part of this PR.

PS: This also adds JS docs for the IC object API.


lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/guard_function/rust.rs
```diff
@@ -4,15 +4,16 @@ use quote::quote;
...
-            let js_guard_fn_return_value =
-                unwrap_boa_result(boa_context.eval_script(boa_engine::Source::from_bytes(#function_call)), &mut boa_context);
+            let js_guard_fn_return_value = boa_context
+                .eval_script(boa_engine::Source::from_bytes(#function_call))
+                .unwrap_or_trap(&mut boa_context);
 
             match js_guard_fn_return_value.try_from_vm_value(&mut *boa_context) {
                 Ok(return_value) => return_value,
-                Err(e) => ic_cdk::api::trap(&format!("Uncaught TypeError: {}",&e.0))
+                Err(e) => ic_cdk::api::trap(&format!("Uncaught {}", &e.0)),
```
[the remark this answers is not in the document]
dansteren: Guard functions return a `Result<(), String>`. Can we discuss if the `unwrap_or_trap` is actually necessary? I've updated the function so that it doesn't call trap, it just returns an `Err(string)` instead which I think is sufficient.

src/compiler/typescript_to_rust/azle_generate/src/body/ic_object/functions/async_call/promise_fulfillment.rs
```diff
@@ -1,52 +1,96 @@
...
+                    &[call_result_js_value],
+                    &mut *boa_context,
+                )
+                .unwrap_or_trap(&mut *boa_context);
 
             let main_promise = PROMISE_MAP_REF_CELL.with(|promise_map_ref_cell| {
                 let promise_map = promise_map_ref_cell.borrow().clone();
 
-                let main_promise = promise_map.get(&uuid).unwrap();
+                let main_promise = promise_map.get(&uuid).unwrap_or_else(|| {
```
lastmjs, line 79: Should this not use `unwrap_or_trap`?

src/compiler/typescript_to_rust/azle_generate/src/body/ic_object/functions/async_call/promise_fulfillment.rs
```diff
@@ -1,52 +1,96 @@
...
-            let mut boa_context = box_context_ref_cell.borrow_mut();
+        BOA_CONTEXT_REF_CELL.with(|boa_context_ref_cell| {
+            let mut boa_context = boa_context_ref_cell.borrow_mut();
 
             let call_result_js_value = match call_result {
                 Ok(value) => {
-                    let js_value = value.try_into_vm_value(&mut *boa_context).unwrap();
+                    let js_value = match value.try_into_vm_value(&mut *boa_context) {
+                        Ok(js_value) => js_value,
+                        Err(vmc_err) => {
```
lastmjs, line 10: What is vmc_err? Can we just call it err?

lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/body/ic_object/functions/async_call/promise_fulfillment.rs
```diff
@@ -1,52 +1,96 @@
...
+                                .unwrap_or_trap(&mut *boa_context);
 
-                    let canister_result_js_object = boa_engine::object::ObjectInitializer::new(&mut *boa_context)
-                        .property(
-                            "Ok",
-                            js_value,
-                            boa_engine::property::Attribute::all()
-                        )
-                        .build();
+                            return;
```
lastmjs, line 20: Let's just return this value directly, not return;

lastmjs (changes requested):

lastmjs (approved):

PR #1110, opened 2023-07-18
fix composite queries

bdemann:


bdemann:
Jordan did this and merged it into main today. I don't know the commit, its the one that happened just before I closed this ticket

PR #1120, opened 2023-07-26
Replaced CdkActTryFromVmValueErrors with JsErrors

### LINKED ISSUE #1119: Replace CdkTryFromVmValueErrors with JsErrors
Every time that we try to convert from a VM value into Rust, it means that we were just in a JS context and are thus able to throw an exception up to that JS layer if anything in the conversion goes wrong. That means we shouldn't create intermediate string representations of the errors, but should instead create native JsErrors. We encounter the same problem in Kybra.

Relates to https://github.com/demergent-labs/kybra/issues/392

dansteren:
Closes #1119


src/compiler/typescript_to_rust/azle_generate/src/vm_value_conversion/try_from_vm_value_impls/basic.rs
```diff
@@ -170,25 +213,34 @@ pub fn generate() -> proc_macro2::TokenStream {
                             }
                         };
 
+                        // TODO: Consider turning these inner errors into custom errors.
```
lastmjs, line 216: Make an issue for this

src/compiler/typescript_to_rust/azle_generate/src/vm_value_conversion/try_from_vm_value_impls/numeric.rs
```diff
@@ -1,195 +1,282 @@
...
-        impl CdkActTryFromVmValue<f64, &mut boa_engine::Context<'_>> for boa_engine::JsValue {
+        impl CdkActTryFromVmValue<f64, boa_engine::JsError, &mut boa_engine::Context<'_>>
+            for boa_engine::JsValue
+        {
             fn try_from_vm_value(
                 self,
                 context: &mut boa_engine::Context,
-            ) -> Result<f64, CdkActTryFromVmValueError> {
-                Ok(self.as_number().ok_or_else(|| "TypeError: Value is not of type 'float64'")?)
+            ) -> Result<f64, boa_engine::JsError> {
```
lastmjs, line 9: Consider using JsResult everywhere

lastmjs (changes requested):

src/compiler/typescript_to_rust/azle_generate/src/vm_value_conversion/try_from_vm_value_impls/basic.rs
dansteren: Done: https://github.com/demergent-labs/azle/issues/1121

PR #1146, opened 2023-08-03
Added check to ensure init and post-upgrade params are of same length

### LINKED ISSUE #1095: Init and post upgrade params must be the same
Right now there is a requirement by at least us (maybe the protocol, seems to be up in the air) to ensure that the init params and the post upgrade params are the same. We should ensure that there is a nice compile-time message if this is not the case.

Ideally the error message should print out a nice message like:

```plaintext
error: params for $init and $postUpgrade must be exactly the same
  --> ./src/index.ts:4:17
   |
 4 |   export function init(p1: boolean, p2: string, p3: int32): void {
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ these params
...
 9 |   export function postUpgrade(p1: int32, p2: boolean): void {
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^ and these params do not match
   |
help: update the params for either $init or $postUpgrade to match the other. E.g.:
   |
 4 |   export function init(p1: boolean, p2: string, p3: int32): void {
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ copy these params
...
 9 |   export function postUpgrade(p1: boolean, p2: string, p3: int32): void {
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ to here
   |
```

- [x] move the to string implementation into DidVisitor or something like that, so that we can use it generally on any IDL.Type
- [x] Let's simplify the _azleInitAndPostUpgradeIdlTypes...it seems like if you just split those into two separate variables, that we could simplify the visitor implementation...just make it _azleInitAndPostUpgradeIdlTypeArgs
- [x] Definitely get rid of the init and post_upgrade type annotation things
- [x] Make sure that none of the system methods can be set more than once
- [x] Any other appropriate error checking

dansteren:
This adds a compilation-time check to ensure that the params for init and post-upgrade are of the same length. If they differ in length then an error is returned to the user. If they are of the same length then no error occurs.

This does not check they types of the params. Checking the types will require deep knowledge of each type reference, including which file it's from etc. and is not tackled as part of this PR. See https://github.com/demergent-labs/azle/issues/1095#issuecomment-1663364443 for more information.

Closes #1095

### DIFF
--- a/examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
+++ b/examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
@@ -1,0 +1,68 @@
+import { execSync } from 'child_process';
+import { Test } from 'azle/test';
+
+interface ExecSyncError extends Error {
+    status: number;
+    signal: string;
+    output: Buffer[];
+    pid: number;
+    stdout: Buffer;
+    stderr: Buffer;
+}
+
+export function getTests(): Test[] {
+    return [
+        {
+            name: 'Using mismatched params in $init and $postUpgrade causes a compilation-time error',
+            test: async () => {
+                try {
+                    execSync(`npx azle canister`, { stdio: 'pipe' });
+
+                    return {
+                        Err: 'Expected the build to fail but it succeeded'
+                    };
+                } catch (e: any) {
+                    let stdErr = (e as ExecSyncError).stderr.toString();
+
+                    if (stdErr.includes(expectedError)) {
+                        return { Ok: true };
+                    } else {
+                        return {
+                            Err: unexpectedErrorMessage(expectedError, stdErr)
+                        };
+                    }
+                }
+            }
+        }
+    ];
+}
+
+const expectedError = `error: params for $init and $postUpgrade must be exactly the same
+  --> ./src/index.ts:4:17
+   |
+ 4 |   export function init(p1: boolean, p2: string, p3: int32): void {
+   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ these params
+...
+ 9 |   export function postUpgrade(p1: int32, p2: boolean): void {
+   |                              ^^^^^^^^^^^^^^^^^^^^^^^^ and these params do not match
+   |
+help: update the params for either init or postUpgrade to match the other. E.g.:
+   |
+ 4 |   export function init(p1: boolean, p2: string, p3: int32): void {
+   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ copy these params
+...
+ 9 |   export function postUpgrade(p1: boolean, p2: string, p3: int32): void {
+   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ to here
+   |`;
+
+function unexpectedErrorMessage(expectedError: string, stdErr: string) {
+    return `>  The output from std err does not contain the expected error
+
+Expected Error:
+
+${expectedError}
+
+ActualError:
+
+${stdErr}`;
+}
--- a/examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
+++ b/examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
@@ -16,29 +16,18 @@ export function getTests(): Test[] {
             name: 'Using mismatched params in $init and $postUpgrade causes a compilation-time error',
             test: async () => {
                 try {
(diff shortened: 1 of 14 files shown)


examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+            name: 'Using mismatched params in $init and $postUpgrade causes a compilation-time error',
+            test: async () => {
+                try {
+                    execSync(`npx azle canister`, { stdio: 'pipe' });
+
+                    return {
+                        Err: 'Expected the build to fail but it succeeded'
+                    };
+                } catch (e: any) {
+                    let stdErr = (e as ExecSyncError).stderr.toString();
```
lastmjs, line 25: const

examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
```diff
@@ -0,0 +1,69 @@
...
+    stderr: Buffer;
+}
+
+export function getTests(): Test[] {
+    return [
+        {
+            name: 'Using mismatched params in $init and $postUpgrade causes a compilation-time error',
+            test: async () => {
+                try {
+                    execSync(`npx azle canister`, { stdio: 'pipe' });
```
lastmjs, line 19: dfx build

src/compiler/typescript_to_rust/azle_generate/src/canister_method/errors/mismatched_post_deploy_params.rs
```diff
@@ -0,0 +1,220 @@
...
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MismatchedPostDeployParams {
+    pub origin: String,
+    pub line_number: usize,
+    pub source: String,
+    pub ranges: Vec<(usize, usize)>,
+}
+
+impl MismatchedPostDeployParams {
```
lastmjs, line 40: We need this to work if init and post_upgrade are in separate files

src/compiler/typescript_to_rust/azle_generate/src/canister_method/errors/mismatched_post_deploy_params.rs
```diff
@@ -0,0 +1,220 @@
...
+    let b_start_byte_pos = get_start_byte_pos(b);
+
+    if a_start_byte_pos.lt(&b_start_byte_pos) {
+        (a, b)
+    } else {
+        (b, a)
+    }
+}
+
+fn get_start_byte_pos(sm_an_fn_decl: &SmAnFnDecl) -> BytePos {
```
lastmjs, line 146: Get rid of the snippets, or comment them out. Let's just keep it simple

lastmjs (changes requested):
Let's rework this using the ACT nodes and no the AST. Check the length of the params.

All AST stuff should be removed I am thinking.

lastmjs:
Make sure to put the test in test.yml

src/compiler/typescript_to_rust/azle_generate/src/canister_method/mod.rs
lastmjs, line 1: In this file, let's try to pull apart validation and retrieving the canister method AST nodes and creating the ACT nodes.

The end goal of this is to have more loosely-coupled code, and to not have to reretrieve the init and post_upgrade AST nodes for the param check.

lastmjs (changes requested):

examples/compiler_errors/mismatched_init_and_post_upgrade_params/test/tests.ts
dansteren: I'm going to use `dfx deploy` as that also runs `dfx canister create`.

src/compiler/typescript_to_rust/azle_generate/src/canister_method/errors/mismatched_post_deploy_params.rs
dansteren: Done in c1aea753d74f473388ac09c48f1a244421af856a

src/compiler/typescript_to_rust/azle_generate/src/canister_method/errors/mismatched_post_deploy_params.rs
dansteren: Done by making the error message very simple instead. See c1aea753d74f473388ac09c48f1a244421af856a

PR #1227, opened 2023-09-12
Manual examples

bdemann:


examples/composite_queries/canisters/canister1/index.ts
```diff
@@ -0,0 +1,89 @@
+import { ic, Manual, nat, Principal, query, Service, text, update } from 'azle';
+import Canister2 from '../canister2';
+
+// Composite query calling a query
```
lastmjs: I think this comment should be moved to its appropriate method if I remember correctly

lastmjs (approved):

Issue #1209, opened 2023-09-07
Get guard functions working with Runtimize Architecture


dansteren:
This still needs refinement...

BUT... it will likely follow what we do for the canister methods of using the decorators to put stuff on the global object and then reading that in Rust.

> [!IMPORTANT]
> Keep in mind that guard_functions are really only supported on `query` and `update` methods. We went way overboard before.

Additionally, make sure to update the `guard_functions` example.

PR #1243, opened 2023-09-19
Guard functions

lastmjs:
Closes https://github.com/demergent-labs/azle/issues/1209


examples/guard_functions/src/guards.ts
```diff
@@ -54,38 +35,19 @@ class CustomError extends Error {
...
-    console.log('returnNonGuardResultObject called');
-    // @ts-ignore
-    return { badProp: 'Something other than a guard result' };
-}
-
-export function returnNonNullOkValue(): GuardResult {
-    console.log('nonNullOkValue called');
-    // @ts-ignore
-    return { Ok: 'Something other than null' };
+    return 'Upgrades to this canister are disabled';
```
dansteren, line 47: Shouldn't this `throw`?

examples/guard_functions/src/invalid.ts
dansteren, line 1: It seems like many of these edge cases aren't valid with the new syntax. However, I wonder if there are some that still are, or if we have even different edge cases. I like that this file gets us thinking about all the ways users might do something wrong. For example, what if they did `@query([], bool, { guard: someInvalidIdentifier })`, what error would they get? I just want to make sure that users get direct, informative error messages when they do something wrong, so I would suggest that we modify this file for the scenarios that make sense with our new syntax, or that we create an issue for improving these errors later.

src/compiler/typescript_to_rust/azle_generate_rearchitecture/src/main.rs
```diff
@@ -104,11 +105,20 @@ fn main() -> Result<(), String> {
...
+            let (guard_attribute, guard_function) = if let Some(guard_name) = &canister_method.guard_name
+            {
+                get_guard_token_stream(guard_name)
+            } else {
+                (quote!(), quote!())
+            };
+
             quote! {
-                #[ic_cdk_macros::query(manual_reply = true, composite = #is_composite)]
+                #[ic_cdk_macros::query(manual_reply = true, composite = #is_composite #guard_attribute)]
```
dansteren, line 116: Is there supposed to be comma between composite and #guard_attribute? I'm sure you did it right but it caught my eye so just double check.

dansteren (approved):
Other than the comments I left it looks good.

examples/guard_functions/src/guards.ts
lastmjs: Not needed I'll just delete it in my next PR

examples/guard_functions/src/invalid.ts
lastmjs: I'm just not sure this makes sense, an invalid identifier would show up as a typescript or javascript error, we don't do many of the things we used to do so I would say it's more likely not our concern.

Generally we want errors to show nicely but it's so different now.

src/compiler/typescript_to_rust/azle_generate_rearchitecture/src/main.rs
lastmjs: It's correct, the comma is in the guard_attribute

PR #1338, opened 2023-10-04
Candid keywords

bdemann:


examples/candid_types/.gitignore
lastmjs, line 1: This example should be named `candid_keywords`

examples/candid_types/test/tests.ts
```diff
@@ -0,0 +1,43 @@
+import { Test } from 'azle/test';
+import { _SERVICE } from './dfx_generated/candid_types/candid_types.did';
+import { ActorSubclass } from '@dfinity/agent';
+
+export function getTests(candid_canister: ActorSubclass<_SERVICE>): Test[] {
+    return [
+        {
+            name: 'test query with keyword name',
+            test: async () => {
+                await candid_canister.opt();
```
lastmjs, line 10: Why not get the result and test that it's correct?

examples/candid_types/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "scripts": {
+        "pretest": "ts-node --transpile-only --ignore=false test/pretest.ts",
+        "test": "ts-node --transpile-only --ignore=false test/test.ts"
+    },
+    "dependencies": {
+        "azle": "0.17.1"
+    },
+    "devDependencies": {
+        "@dfinity/agent": "0.11.1",
```
lastmjs, line 10: These dependencies should be up-to-date

lastmjs (changes requested):

examples/candid_types/test/tests.ts
bdemann: Because if it fails it will fail before we get the result, the result is not necessary

lastmjs (approved):

PR #1373, opened 2023-10-12
refactor js rearchitecture code

bdemann:


src/lib/ic/stable_64_read.ts
lastmjs, line 1: In all of the IC objects, we should get rid of the `BigInt` runtime call and just cast to `bigint` if possible

src/lib/canister_methods/index.ts
lastmjs, line 1: Move executeMethod into its file, move Manual and types into their files

src/lib/canister_methods/pre_upgrade.ts
```diff
@@ -5,7 +5,7 @@ export function preUpgrade(
     callback: () => void | Promise<void>
 ): CanisterMethodInfo<[], Void> {
     const finalCallback = (...args: any[]) => {
-        executeMethod('preUpgrade', args, callback, [], Void, false);
+        callback();
```
lastmjs, line 8: I think this can be reduced to just passing the callback in, get rid of the finalCallback concept

src/lib/canister_methods/heartbeat.ts
```diff
@@ -1,11 +1,12 @@
 import { CanisterMethodInfo, executeMethod, isAsync } from '.';
+import { ic } from '../ic';
 import { Void } from '../candid/types/primitive/void';
 
 export function heartbeat(
     callback: () => void | Promise<void>
 ): CanisterMethodInfo<[], Void> {
-    const finalCallback = (...args: any[]) => {
-        executeMethod('heartbeat', args, callback, [], Void, false);
+    const finalCallback = () => {
```
lastmjs, line 8: This can be simplified, I don't think we need the finalCallback concept

src/lib/candid/index.ts
lastmjs, line 1: Pull typemapping into its own file

lastmjs, line 1: CandidType should also be in its own file

src/lib/candid/to_Idl.ts
lastmjs, line 1: Rename function with all lowercase

src/lib/candid/to_Idl.ts
```diff
@@ -0,0 +1,45 @@
+import { IDL } from '@dfinity/candid';
```
lastmjs, line 1: See if it's nice to combine all of these functions into just toIdl

src/lib/candid/to_Idl.ts
lastmjs, line 1: Rename file to `to_idl` all lowercase

src/lib/candid/types/constructed/index.ts
```diff
@@ -8,19 +8,18 @@ export * from './tuple';
 export * from './variant';
 export * from './vec';
 
-type CandidMap = { [key: string]: CandidType };
-type IdlMap = { [key: string]: IDL.Type<any> };
+export type CandidMap = { [key: string]: CandidType };
```
lastmjs, line 11: Moving everything but the exports into their own files

src/lib/candid/serde/index.ts
```diff
@@ -18,24 +18,18 @@ import { CandidType, Parent, toIDLType } from '../../candid';
  * @returns candid bytes
  */
 export function encode(
```
lastmjs, line 20: Pull these functions out into their own files

src/lib/candid/did_visitor/index.ts
lastmjs, line 1: Rename top level folder to `did_file`, create visitors directory just like in serde

lastmjs, line 1: Move things in here into their own files, exports only from this file is possible

lastmjs (changes requested):

src/lib/candid/to_Idl.ts
bdemann: We ought not combine the three functions into one, but we can combine toParamIdl and toReturnIdl into one function, that will also make that function more generalized so we can use it to simplify encode and decode multiple.

The reason we want toIdl and what will be called toIdlArray is because one of them will return an array always (regardless of how many items are passed) and one of them will return a single value always

src/lib/candid/did_file/visitor.ts/escapeCandidKeywords.ts
lastmjs, line 1: This file should be snake case

src/lib/candid/did_file/visitor.ts/extract_candid.ts
```diff
@@ -0,0 +1,16 @@
+import { CandidDef, CandidTypesDefs } from '.';
```
lastmjs, line 1: Should not import from index

src/lib/ic/call.ts
lastmjs, line 1: Check everywhere for type errors

lastmjs (changes requested):

src/lib/candid/did_file/visitor/visit/func.ts
lastmjs, line 1: Check everywhere to make sure imports come from their files directly, not from an export *

lastmjs (changes requested):

bdemann:
> serde visitors should look like did_file visitor, each visitor should have its own file in a folder called visit

Done

lastmjs (approved):

nezort11:
@lastmjs 

Hi, after this change made 3 weeks ago:

image

The return type of `StableBTreeMap` is always `any`.

image

if I comment out this 3 lines of code, it is back to normal:

image

PR #1402, opened 2023-10-20
add primitive property tests

bdemann:


bdemann:
closing in favor of breaking each example into their own examples

PR #1420, opened 2023-10-23
add test support for NaNs and infinities

bdemann:


property_tests/are_equal.ts
lastmjs, line 1: The name of this file should probably match the function name more closely for floats

property_tests/tests/float32/test/test.ts
```diff
@@ -25,7 +26,11 @@ const Float32TestArb = fc
 
         const paramsCorrectlyOrdered = paramNames
             .map((paramName, index) => {
-                return `if (${paramName} !== ${paramSamples[index]}) throw new Error('${paramName} is incorrectly ordered')`;
+                let areFloat32sEqual = areFloatsEqual(
```
lastmjs, line 29: const not let

property_tests/tests/float64/test/test.ts
```diff
@@ -33,7 +34,11 @@ const Float64TestArb = fc
 
         const paramsCorrectlyOrdered = paramNames
             .map((paramName, index) => {
-                return `if (${paramName} !== ${paramSamples[index]}) throw new Error('${paramName} is incorrectly ordered')`;
+                let areFloat64sEqual = areFloatsEqual(
```
lastmjs, line 37: const not let

lastmjs (changes requested):

PR #1422, opened 2023-10-24
Add principal property tests

dansteren:


property_tests/arbitraries/candid/reference/principal_arb.ts
```diff
@@ -0,0 +1,11 @@
+import fc from 'fast-check';
+import { Principal } from '@dfinity/principal';
+
+export const PrincipalArb = fc
+    .uint8Array({
+        minLength: 29,
+        maxLength: 29,
+        min: 0,
```
lastmjs, line 8: I don't think we need the min and max

lastmjs (changes requested):

property_tests/arbitraries/candid/reference/principal_arb.ts
dansteren: removed

PR #1423, opened 2023-10-25
Property tests recursive

bdemann:


bdemann:
We should eventually close this PR. But there are some lingering notes from Jordan's unsubmitted review that I need to get before we do so. We will not be merging these changes into main via this PR. Instead Dan and I will be merging a bunch of changes to https://github.com/demergent-labs/azle/tree/property_test_candid_types (I'll update with the PR link when I have it)

lastmjs (changes requested):

PR #1431, opened 2023-10-30
update arbitries to have their own equals method

bdemann:


property_tests/arbitraries/candid/constructed/vec_arb.ts
```diff
@@ -11,23 +10,28 @@ import { Nat16Arb } from '../primitive/nats/nat16_arb';
...
+            arraysAreEqual(a, b, sample[0]?.equals ?? deepEqual);
+        return {
             value: sample.map((sample) => sample.value),
             src: {
                 candidType: `Vec(${src.src.candidType})`,
-                imports: new Set(src.src.imports).add('Vec')
-            }
-        })
-    );
+                imports: new Set(src.src.imports).add('Vec'),
```
dansteren, line 29: Let's try to not do `.add` so check for all the places you're using it. I won't comment on this again.

property_tests/are_equal/opt.ts
```diff
@@ -1,40 +1,48 @@
...
+function stringifyBigInts(_key: any, value: any) {
+    if (typeof value === 'bigint') {
+        return value.toString() + 'n';
     }
+    return value;
+}
 
-    return value1 === value2;
+// NOTE: This is a little buggy but seems to work for opt (bigints are "123n" instead of 123n)
+function valueToSrc(value: any): string {
```
dansteren, line 46: Remove this if it's not used.

property_tests/tests/null/test/test.ts
```diff
@@ -5,48 +5,73 @@ import { JsFunctionNameArb } from '../../../arbitraries/js_function_name_arb';
...
+    const areAllNull = paramNames.reduce((acc, paramName) => {
+        return `${acc} && ${paramName} === null`;
+    }, 'true');
+
+    const allNullCheck = `if (!(${areAllNull})) throw new Error("Not all of the values were null")`;
+
+    return `
+        ${allNullCheck}
+
+        return ${returnNull.value};
```
dansteren, line 54: This seems not correct. I think you want the string representation of the null value. Put a todo in so that we don't forget about this. We just want the code to say what we're meaning.

dansteren (changes requested):

dansteren (approved):
Awesome! :star_struck:

PR #1432, opened 2023-10-31
Add property tests for Tuples

dansteren:


property_tests/tests/tuple/test/test.ts
```diff
@@ -0,0 +1,152 @@
...
+            const paramName = `param${index}`;
+            const fieldsCount = tuple.value.length;
+
+            const paramIsArray = `Array.isArray(${paramName})`;
+            const paramHasCorrectNumberOfFields = `${paramName}.length === ${fieldsCount}`;
+            const throwError = `throw new Error('${paramName} must be a Tuple');`;
+
+            return `if (!(${paramIsArray} && ${paramHasCorrectNumberOfFields})) ${throwError}`;
+        })
+        .join('\n        ');
```
bdemann, line 69: I like tabs and your attention to detail, but I don't think we should do this

bdemann (changes requested):

PR #1442, opened 2023-11-07
Property test deep equal

bdemann:


property_tests/tests/float32/test/test.ts
```diff
@@ -66,11 +66,7 @@ function generateBody(
     );
     const paramsCorrectlyOrdered = paramNames
         .map((paramName, index) => {
-            const areFloat32sEqual = areFloatsEqual(
-                paramName,
-                paramLiterals[index]
-            );
-            return `if (!(${areFloat32sEqual})) throw new Error('${paramName} is incorrectly ordered')`;
+            return `if (!deepEqual(${paramName}, ${paramLiterals[index]})) throw new Error('${paramName} is incorrectly ordered')`;
```
lastmjs, line 69: Abstract this away since it should be the same check for all tests, write this down.

property_tests/tests/func/test/test.ts
```diff
@@ -97,7 +98,8 @@ function generateBody(
             const methodNameIsCorrect = `${paramName}[1] === '${methodName}'`;
             const throwError = `throw new Error('${paramName} is incorrectly ordered')`;
 
-            return `if (!(${principalValueIsCorrect} && ${methodNameIsCorrect})) ${throwError}`;
+            const debug = `console.log(${paramName})\nconsole.log(${func.src.valueLiteral})\n`;
```
lastmjs, line 101: Remove debug

property_tests/tests/opt/test/test.ts
```diff
@@ -73,18 +70,14 @@ function generateBody(
 
     const areParamsCorrectlyOrdered = paramNames
         .map((paramName, index) => {
-            return `if (!${createAreOptsEqualCodeUsage(
-                paramName,
-                paramLiterals[index]
-            )}) throw new Error('${paramName} is incorrectly ordered')`;
+            return `if (false && !deepEqual(${paramName}, ${paramLiterals[index]})) throw new Error('${paramName} is incorrectly ordered')`;
```
lastmjs, line 73: I think we might want to remove this false && thing

lastmjs (changes requested):

property_tests/tests/float32/test/test.ts
bdemann: What if instead of writing it down, I just did it real quick? @lastmjs

lastmjs: I think I'd get a kick out of that

lastmjs (approved):

PR #1447, opened 2023-11-09
refactor CandidType and Serializable

lastmjs:


src/lib/stable_structures/stable_b_tree_map.ts
```diff
@@ -1,24 +1,24 @@
...
 export interface Serializable {
     toBytes: (data: any) => Uint8Array;
     fromBytes: (bytes: Uint8Array) => any;
 }
 
 export function StableBTreeMap<Key = any, Value = any>(
-    keyType: Partial<Serializable>,
-    valueType: Partial<Serializable>,
+    keySerializable: Serializable,
+    valueSerializable: Serializable,
```
dansteren, line 13: I'd recommend just `key` and `value`. The type tells you it's serializable. Just an idea though so just resolve this comment if you like it as is.

src/lib/stable_structures/stable_b_tree_map.ts
```diff
@@ -1,24 +1,24 @@
...
 import { nat8 } from '../candid/types/primitive/nats/nat8';
-import { encode, decode } from '../candid/serde';
+import { decode } from '../candid/serde';
 
 export interface Serializable {
     toBytes: (data: any) => Uint8Array;
     fromBytes: (bytes: Uint8Array) => any;
 }
 
 export function StableBTreeMap<Key = any, Value = any>(
```
dansteren, line 11: Now that this function can throw, I think it's important to document that as a JS doc comment. I'd use the [`@throw`](https://jsdoc.app/tags-throws) tag. Now would also be a good time to add more documentation explaining how to use StableBTreeMap, for example maybe adding `@typeparam` tags to explain the `Key`, and `Value` generics.

src/lib/stable_structures/stable_b_tree_map.ts
```diff
@@ -1,24 +1,24 @@
...
     memoryId: nat8
 ) {
     if (globalThis._azleIc !== undefined) {
         globalThis._azleIc.stableBTreeMapInit(memoryId);
     }
 
-    isSerializable(keyType);
-    isSerializable(valueType);
+    isSerializable(keySerializable);
+    isSerializable(valueSerializable);
```
dansteren, line 21: An opinionated naming preference thing here (i.e. feel free to ignore), `isSerializable` to me doesn't say it will throw or assert so I'd go with something more like `assertIsSerializable` or `ensureIsSerializable`. Just a thought.

dansteren (approved):
I really think the throw should be documented before this is merged. That being said, it wasn't technically part of this PR, so I've gone ahead and approved it anyway. I just don't want us to forget.

And for the other stuff just choose what you think is best.

src/lib/stable_structures/stable_b_tree_map.ts
dansteren: ah, figured out what it is. `isSerializable` to me says it will return a boolean. I think that's my issue. So I read it as you checking whether the thing was serializable, and then not doing anything with the result. So I think a different name wouldn't conflict with that convention.

lastmjs:
I'm still working on StableBTreeMap so this kind of documentation isn't really what I was focusing on

src/lib/stable_structures/stable_b_tree_map.ts
lastmjs: key and value are already used inside of each method though

src/lib/stable_structures/stable_b_tree_map.ts
lastmjs: Documentation is good but nothing really throws here, isSerializable is just a type assertion it's not really meant to do anything at runtime, though it will.

This is a pattern that comes up often, and generally I think we don't want to rely on runtime type checking. The throw is just for the compiler.

src/lib/stable_structures/stable_b_tree_map.ts
lastmjs: This is the convention in TypeScript for these if I'm not mistaken...just what I've always seen

PR #1448, opened 2023-11-09
Typeof tstype

lastmjs:


dansteren (dismissed):
TBH I didn't see the difference between this and your other PR, but both look good. See my other review for feedback.

dansteren (approved):
New changes look good! _(and are much smaller)_

PR #1449, opened 2023-11-09
String serde for bigint

lastmjs:


dansteren (dismissed):
My understanding is that you swapped out the crazy candid encoding for a simpler `toString` serialization for types where it made sense. It might be good to just discuss in the morning to make sure I fully understand, but I didn't see any glaring omissions.

PR #1450, opened 2023-11-16
Stable b tree map performance tests

lastmjs:


src/lib/stable_structures/stable_b_tree_map.ts
```diff
@@ -191,8 +191,8 @@ export function StableBTreeMap<Key = any, Value = any>(
 
             const encodedValues = globalThis._azleIc.stableBTreeMapValues(
                 memoryId,
-                startIndex,
-                length
+                Math.floor(startIndex),
```
bdemann, line 194: This is fine, I'm just curious at what happened that caused you to put it in

bdemann (approved):

dansteren (approved):

PR #1451, opened 2023-11-16
Stable b tree map property tests

lastmjs:


property_tests/arbitraries/candid/reference/service_arb/definition_arb.ts
```diff
@@ -80,30 +93,53 @@ function ServiceMethodArb(
             return {
                 name,
                 imports,
-                typeAliasDeclarations,
+                variableAliasDeclarations: variableAliasDeclarations,
```
dansteren, line 96: This should probably use object shorthand syntax. The linter should really catch things like this though, see https://eslint.org/docs/latest/rules/object-shorthand.

property_tests/tests/stable_b_tree_map/test/contains_key.ts
```diff
@@ -0,0 +1,74 @@
...
+): Test {
+    return {
+        name: `containsKey ${functionName}`,
+        test: async () => {
+            const actor = getActor('./tests/stable_b_tree_map/test');
+
+            const result = await actor[functionName](param0Value, param1Value);
+
+            return {
+                Ok: deepEqual(result, true)
```
dansteren, line 70: Why not `===`? Does this mean that I should always use `deepEquals` for all tests?

property_tests/arbitraries/stable_b_tree_map_arb.ts
```diff
@@ -0,0 +1,30 @@
...
+                max: 254
+            })
+        )
+    )
+    .map(([keySample, valueSample, uniqueIdentifier, memoryId]) => {
+        const name = `stableBTreeMap${uniqueIdentifier}`;
+
+        return {
+            name,
+            body: `let ${name} = StableBTreeMap<${keySample.src.candidTypeAnnotation}, ${valueSample.src.candidTypeAnnotation}>(stableJson, stableJson, ${memoryId});`,
```
dansteren, line 22: hmmm, `body` is an interesting name here. When I was reading it later in the StableBTreeMap method tests it read more like it was something else. Isn't this a constructor, or variable declaration? I just wouldn't think of this as a "body" so maybe there's a better name for this.

property_tests/tests/stable_b_tree_map/test/get.ts
```diff
@@ -0,0 +1,78 @@
...
+            tests: [test]
+        };
+    });
+
+function generateBody(
+    stableBTreeMapName: string,
+    stableBTreeMapBody: string
+): string {
+    return `
+        ${stableBTreeMapBody}
```
dansteren, line 53: Calling this `stableBTreeMapBody` doesn't seem quite right. It's creating the map, so I'd lean more to something more like constructor. It feels like it's going to be like the following two lines, but then isn't. I'm not explaining well though so let's talk about this IRL if you'd like.

dansteren (commented):
Well done! This is a lot of work and a good job adapting to all our merge conflicts :grimacing:  (sorry).

I'd be fine if we merged it as is, but I think it's still a work in progress. I'm sensing not quite the abstraction on the StableBTreeMap arb as a few of my comments note. That's the biggest architectural problem I see. That being said, I realize it's a step forward so I'm fine merging it if you'd like. So, I'm "commenting", rather than "approving" or "requesting changes".

PR #1454, opened 2023-11-17
Property test arbitrary depth variant

bdemann:


property_tests/arbitraries/candid/constructed/record_arb/base.ts
```diff
@@ -0,0 +1,110 @@
...
+import { JsFunctionNameArb } from '../../../js_function_name_arb';
+import { Record } from './index';
+
+type Field = [string, CandidMeta<CandidType>];
+
+export function RecordArb(candidTypeArb: fc.Arbitrary<CandidMeta<CandidType>>) {
+    return fc
+        .tuple(
+            UniqueIdentifierArb('typeDeclaration'),
+            fc.uniqueArray(fc.tuple(JsFunctionNameArb, candidTypeArb), {
```
lastmjs, line 15: Is JsFunctionNameArb unique?

lastmjs: Is that what uniqueArray is doing somehow? What happens if there are two JsFunctionNameArbs with the same name? Seems like we should be using UniqueIdentifierArb

property_tests/arbitraries/candid/constructed/variant_arb/base.ts
```diff
@@ -0,0 +1,125 @@
...
+
+function VariantFieldsArb(
+    candidTypeArb: fc.Arbitrary<CandidMeta<CandidType>>
+): fc.Arbitrary<Field[]> {
+    return fc.uniqueArray(fc.tuple(JsFunctionNameArb, candidTypeArb), {
+        selector: (entry) => entry[0],
+        minLength: 1
+        // Although no minLength is technically required (according to the
+        // spec), the DFX CLI itself currently errors out trying to pass
+        // an empty object.
```
lastmjs, line 18: Is this still a problem? We should open an issue somewhere

property_tests/arbitraries/candid/candid_arb.ts
```diff
@@ -1,16 +1,18 @@
 import fc from 'fast-check';
-import { deepEqual } from 'fast-equals';
 import { CandidType } from './candid_type_arb';
 
 // TODO we're thinking that Candid is not the best name for this. What is better?
-export type CandidMeta<T extends CandidType> = {
+export type CandidMeta<T extends CandidType, E = T> = {
     value: T;
```
lastmjs, line 6: Can we think of a better name for this, somehow to distinguish these values? It's just not clear what the value is and why it's different than the expected value...do you think this is important?

For example should we say agentArgumentValue and agentResponseValue or something?

property_tests/tests/tuple/test/test.ts
```diff
@@ -94,12 +95,19 @@ function generateBody(
     `;
 }
 
+function replacer(key: any, value: any) {
```
lastmjs, line 98: What is this replacer for? is there a JSON.stringify somewhere? We should use the replacer and reviver I've created now, perhaps we should open an issue for this or comment it in here

lastmjs (changes requested):

property_tests/arbitraries/candid/candid_arb.ts
lastmjs: Change these to argumentValue and expectedResponseValue

property_tests/tests/tuple/test/test.ts
lastmjs: Remove this

property_tests/arbitraries/candid/constructed/variant_arb/base.ts
bdemann: We should also open an issue for the tuple inconsistency we found

property_tests/arbitraries/candid/constructed/record_arb/base.ts
bdemann: Unique Identifier Arb is going to make sure that globally all of the names are unique, for this case we don't need to them to be globally unique, we just need them to be locally unique to this particular record, another record could have fields of the same name and that would be okay. uniqueArray is what we want here.

lastmjs (approved):

PR #1458, opened 2023-11-28
Enable ESLint

dansteren:
Given that our codebase is now primarily JS/TS, it's time to add a linter. This PR adds ESLint to the repo with the default recommendations (_those in [this list](https://eslint.org/docs/latest/rules/) with a ✅ and [these typescript rules](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts)_). It has the these benefits:

- imports are automatically sorted
- it is compatible with Prettier
- linter run on commit

> [!WARNING]  
> Some rules in .eslintrc have been disabled to get the linter to pass. This is tech debt and we should go through and address these and resolve the conflicts.

We can consider adding more rules, extending a popular style guide (like Airbnb's), or adding some things like https://github.com/import-js/eslint-plugin-import/ and https://github.com/gajus/eslint-plugin-jsdoc. There's lots we could do to improve/refine this.  This is just a starting point.

Other rules we might want to add:
* [object-shorthand](https://eslint.org/docs/latest/rules/object-shorthand)
* [padding-line-between-statements](https://eslint.org/docs/latest/rules/padding-line-between-statements)


bdemann:
Closing in favor of https://github.com/demergent-labs/azle/pull/1733 and https://github.com/demergent-labs/azle/pull/1734

PR #1466, opened 2023-12-04
Test: Update Canister to accept update methods

dansteren:


property_tests/arbitraries/canister_arb.ts
```diff
@@ -60,3 +69,7 @@ function generateSourceCode(queryMethods: QueryMethod[]) {
         });
     `;
 }
+
+function EmptyArrayArb<T>(): fc.Arbitrary<T[]> {
```
lastmjs, line 73: Let's try to remove this concept and just use `fc.constant([])`

lastmjs (changes requested):

lastmjs (approved):

PR #1468, opened 2023-12-04
Candid type object

bdemann:


bdemann:
this isn't as helpful as I thought it was going to be

PR #1472, opened 2023-12-06
Test rec

bdemann:


bdemann:
Close in favor of https://github.com/demergent-labs/azle/pull/1515

PR #1487, opened 2023-12-08
Console log

lastmjs:


src/lib/globals.ts
```diff
@@ -11,6 +12,19 @@ declare global {
     var _azleGuardFunctions: { [key: string]: () => any };
 }
 
+if (globalThis._azleIc) {
+    globalThis.console = {
+        ...globalThis.console,
+        log: (...args: any[]) => {
+            const jsonStringifiedArgs = args
+                .map((arg) => JSON.stringify(arg, replacer, 4))
+                .join(', ');
```
dansteren, line 21: Both node and chrome don't join with a `,`, they just join with a space...

src/compiler/rust/canister/src/ic/print.rs
```diff
@@ -1,13 +1,19 @@
...
     _this: &CallbackArg,
     args: &[CallbackArg],
 ) -> Result<JSValueRef<'a>, anyhow::Error> {
-    for arg in args {
-        let value = arg.to_js_value()?;
-        ic_cdk::println!("{:?} ", value);
-    }
+    let string_to_print: String = args
+        .get(0)
+        .expect("console.log must have at least one argument")
```
dansteren, line 12: `console.log()` (without params) is valid in JS. Just make sure that still works.

property_tests/tests/stable_b_tree_map/test/test.ts
```diff
@@ -65,8 +65,8 @@ const StableBTreeMapTestArb = fc
                 );
         }),
         {
-            minLength: 1,
-            maxLength: 50
+            minLength: 20,
+            maxLength: 100
```
dansteren, line 69: `defaultArrayConstraints` in azle/property_tests is 20-100, consider using that instead.

property_tests/arbitraries/stable_b_tree_map_arb.ts
```diff
@@ -6,33 +6,76 @@ import { CandidValueAndMetaArb } from './candid/candid_value_and_meta_arb';
...
 
 // TODO there must be a better way to get this type out, the sample will actually run
 export const StableBTreeMap = fc.sample(StableBTreeMapArb)[0];
 export type StableBTreeMap = typeof StableBTreeMap;
+
+function getSerializableArguments(
+    keySample: StableBTreeMap['keySample'],
+    valueSample: StableBTreeMap['valueSample'],
+    keyArgumentInfo: 'STABLE_JSON' | 'CANDID_TYPE_OBJECT',
+    valueArgumentInfo: 'STABLE_JSON' | 'CANDID_TYPE_OBJECT'
```
dansteren, line 62: Seems like this union type should be pulled out rather than duplicated.

Also, `keyArgumentInfo` and `valueArgumentInfo` don't feel informative to me. These seem to be serialization formats. Maybe consider a name more like that.

dansteren (changes requested):

src/lib/globals.ts
lastmjs: Fixed

src/compiler/rust/canister/src/ic/print.rs
lastmjs: Addressed

PR #1498, opened 2023-12-13
Fix Float32Arb precision issue

dansteren:
Floats generated by fc.float may lose precision. Floats that fit in a Float32Array however match the precision of Rust/Candid.

> [!NOTE]
> This is already what we do for float64s.

We recently made changes to account for -0. If this doesn't fix those issues, then we can incorporate those changes with this instead:

```ts
export function Float32ValueArb(): fc.Arbitrary<CandidValues<number>> {
    return SimpleCandidValuesArb(
        fc
            .float32Array({ maxLength: 1, minLength: 1 })
            .map(([sample]) => (sample === 0 ? sample * 0 : sample)),
        floatToSrcLiteral
    );
}
```

If we want to exclude `Infinity` and `NaN` we could also do:

```ts
export function Float32ValueArb(): fc.Arbitrary<CandidValues<number>> {
    return SimpleCandidValuesArb(
        fc
            .float32Array({
                maxLength: 1,
                minLength: 1,
                noDefaultInfinity: true,
                noNaN: true
            })
            .map(([sample]) => (sample === 0 ? sample * 0 : sample)),
        floatToSrcLiteral
    );
}
```

### DIFF
--- a/property_tests/arbitraries/candid/primitive/floats/float32_arb.ts
+++ b/property_tests/arbitraries/candid/primitive/floats/float32_arb.ts
@@ -1,4 +1,4 @@
-import fc from 'fast-check';
+import fc, { sample } from 'fast-check';
 import { floatToSrcLiteral } from '../../to_src_literal/float';
 import { SimpleCandidDefinitionArb } from '../../simple_type_arbs/definition_arb';
 import { SimpleCandidValuesArb } from '../../simple_type_arbs/values_arb';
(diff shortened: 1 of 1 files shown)


dansteren:
We may want to consider adding the `noDefaultInifity` and `noNan` options. I just got an `Infinity` value on one my runs:

image

lastmjs (approved):

lastmjs (approved):

PR #1492, opened 2023-12-11
Test: Added init-method property test

dansteren:
Closes #1471

PR #1512, opened 2023-12-14
Test: Added PostUpgradeArb

### LINKED ISSUE #1497: post_upgrade

dansteren:
Builds off of #1492 to add PostUpgrade tests to our canister methods.

> [!Caution]
> Change the base of this brach to `main` before merging.

Closes #1497

### DIFF
--- a/property_tests/tests/post_upgrade_method/test/generate_init_tests.ts
+++ b/property_tests/tests/post_upgrade_method/test/generate_init_tests.ts
@@ -1,0 +1,34 @@
+import { deepEqual, Named, getActor } from 'azle/property_tests';
+import { CandidValueAndMeta } from 'azle/property_tests/arbitraries/candid/candid_value_and_meta_arb';
+import { CorrespondingJSType } from 'azle/property_tests/arbitraries/candid/corresponding_js_type';
+import { Test } from 'azle/test';
+
+export function generateTests(
+    _functionName: string,
+    params: Named<CandidValueAndMeta<CorrespondingJSType>>[]
+): Test[][] {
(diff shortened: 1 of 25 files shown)


property_tests/tests/post_upgrade_method/test/generate_init_tests.ts
```diff
@@ -0,0 +1,34 @@
...
+        ...params.map((param) => param.el.value.agentResponseValue)
+    ];
+
+    return [
+        [
+            {
+                name: `init method`,
+                test: async () => {
+                    const actor = getActor('./tests/post_upgrade_method/test');
+                    const result = await actor.getPostDeployValues();
```
lastmjs, line 21: I think this needs to be redone. The init method tests should be completely separate from the post upgrade tests. These tests should be almost identical to the actual init method tests, using the same global variables. We do not want the global variables for testing init and post_upgrade to be the same, otherwise you don't have a guarantee that a deploy occurred.

property_tests/tests/post_upgrade_method/test/test.ts
```diff
@@ -0,0 +1,133 @@
...
+                postUpgradeMethod,
+                queryMethods: [getPostDeployValues, ...queryMethods],
+                updateMethods
+            };
+        }
+    );
+
+runPropTests(CanisterArb(CanisterConfigArb));
+
+function generateGetPostDeployValuesCanisterMethod(
```
lastmjs, line 119: Is this how we did the getInitValues in the init method test?

lastmjs (changes requested):

PR #1531, opened 2023-12-27
Test: Added HttpRequestMethodArb

### LINKED ISSUE #1522: http_request
The generated canister might look something like this:

```ts
import {
    Canister,
    HttpRequest,
    HttpResponse,
    None,
    query,
    Record,
    text
} from 'azle';

// #region Variables passed to test
const METHOD = 'GET';
const URL = '/users/7';
const HEADER = [{ name: '', value: '' }];
const I = 0;

const RESPONSE = {
    STATUS_CODE: 204,
    BODY: 'this is the expected return value'
};
// #endregion Variables passed to test

const Token = Record({
    // add whatever fields you'd like
    arbitrary_data: text
});
type Token = typeof Token.tsType;

export default Canister({
    http_request: query([HttpRequest], HttpResponse(Token), (req) => {
        if (req.method !== `${METHOD}`) {
            throw new Error(
                `Unexpected req.method. Expected ${METHOD} but received ${req.method}`
            );
        }
        if (req.url !== `${URL}`) {
            throw new Error(
                `Unexpected req.url. Expected ${URL} but received ${req.url}`
            );
        }
        if (req.headers[I][0] !== `${HEADER[I].name}`) {
            throw new Error(
                `Unexpected header name at index ${I}. Expected ${HEADER[I].name} but received ${req.headers[I][0]}`
            );
        }
        if (req.headers[I][1] !== `${HEADER[I].value}`) {
            throw new Error(
                `Unexpected header value at index ${I}. Expected ${HEADER[I].value} but received ${req.headers[I][1]}`
            );
        }

        return {
            status_code: RESPONSE.STATUS_CODE,
            headers: [],
            body: Buffer.from(RESPONSE.BODY, 'utf-8'),
            streaming_strategy: None,
            upgrade: None
        };
    })
});
```

Resources:
- https://forum.dfinity.org/t/feature-request-map-appropriate-http-request-methods-to-update-calls/4303
- https://github.com/dfinity/icx-proxy/pull/6
- https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec

dansteren:
> [!Caution]
> Change the base of this brach to `main` before merging.

Closes #1522

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,6 +133,7 @@ jobs:
                     "examples/vanilla_js",
                     "property_tests/tests/blob",
                     "property_tests/tests/bool",
+                    "property_tests/tests/canister_methods/http_request",
                     "property_tests/tests/canister_methods/init",
                     "property_tests/tests/canister_methods/post_upgrade",
                     "property_tests/tests/canister_methods/pre_upgrade",
--- a/examples/motoko_examples/http_counter/src/index.ts
+++ b/examples/motoko_examples/http_counter/src/index.ts
@@ -1,63 +1,25 @@
 import {
     blob,
-    bool,
     Canister,
-    Func,
+    HeaderField,
+    HttpRequest,
+    HttpResponse,
     ic,
     init,
     nat,
-    nat16,
     None,
-    Opt,
     query,
     Record,
     Some,
     StableBTreeMap,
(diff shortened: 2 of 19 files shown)


bdemann:
My recollection is that this was closed in favor of doing all of these fixes in here https://github.com/demergent-labs/azle/pull/1533

PR #1533, opened 2023-12-29
Test: Add HTTP request update property tests

### LINKED ISSUE #1532: http_request_update
To perform an http request call the canister needs to expose two endpoints: `http_request` as a query method, and `http_request_update` as an update method. The `http_request` method needs to return an `upgrade` value of `Some(true)` and all other fields are ignored.

We already have tests for `http_request` so it is proposed that this canister's `http_request` method be hardcoded rather than arbitrary.

The generated canister may look like this:

```ts
import {
    Canister,
    HttpRequest,
    HttpResponse,
    nat8,
    None,
    Null,
    query,
    Some,
    update
} from 'azle';

const state: number = 0;

export default Canister({
    http_request: query([HttpRequest], HttpResponse(Null), () => {
        return {
            status_code: 204,
            headers: [],
            body: new Uint8Array(),
            streaming_strategy: None,
            upgrade: Some(true)
        };
    }),
    http_request_update: update(
        [HttpRequest],
        HttpResponse(Null),
        (request) => {
            const expectations = {};
            if (request !== expectations) {
                throw new Error('The http request was different than expected');
            }

            return {
                status_code: 201,
                headers: [
                    ['content-type', 'text/plain'],
                    ['content-encoding', 'gzip']
                ],
                body: Uint8Array.from([
                    31, 139, 8, 0, 55, 2, 27, 98, 0, 3, 43, 45, 72, 73, 44, 73,
                    229, 2, 0, 168, 218, 145, 108, 7, 0, 0, 0
                ]),
                streaming_strategy: None,
                upgrade: None
            };
        }
    ),
    get_state: query([], nat8, () => {
        return state;
    })
});
```

Then we can hit the `http_request` endpoint with various http requests with different methods, urls, headers, and bodies, and then update can effect state changes...

So we also need a way to get the state changes back out... so we could use http_request and have it return the state... but then we'd have to control how it decides to do a query vs update... I think it's better to just leave http_request alone and instead have a second method for querying state.

So the test will be, when we make an arbitrary http request:
- Does the body request param inside of `http_request_update` decode correctly?
- Does it return data correctly?
- Does it update the state?

So tests would look like:
- get the state before
- call http_request
- check that the body is good
- verify the returned response
- check the state again

Resources:
- https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec#upgrade-to-update-calls

dansteren:
Closes #1532


property_tests/tests/canister_methods/http_request/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "scripts": {
+        "test": "ts-node --transpile-only --ignore=false test/test.ts"
```
lastmjs, line 3: this seems fishy, no prepare step?

property_tests/tests/canister_methods/http_request_update/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "scripts": {
+        "test": "ts-node --transpile-only --ignore=false test/test.ts"
```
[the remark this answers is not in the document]
bdemann: none of the property tests have the pretests

property_tests/tests/canister_methods/http_request/package.json
bdemann: none of the property tests have the pretests

property_tests/tests/canister_methods/http_request/dfx.json
```diff
@@ -0,0 +1,16 @@
+{
+    "canisters": {
+        "canister": {
+            "type": "custom",
+            "main": "src/index.ts",
+            "candid": "src/index.did",
+            "build": "npx azle canister",
+            "wasm": ".azle/canister/canister.wasm",
+            "gzip": true,
+            "declarations": {
```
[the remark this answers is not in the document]
bdemann: It doesn't look like it, I will try and get rid of it, and make sure it works

src/lib/server.ts
```diff
@@ -27,38 +28,65 @@ import { IncomingMessageForServer } from 'http';
...
+export const DefaultToken = Record({});
+export function StreamingCallbackHttpResponse<Token extends CandidType>(
+    token: Token
+) {
+    return Record({
+        body: blob,
+        token: Opt(token)
+    });
+}
+type StreamingCallbackHttpResponse<Token> = {
```
lastmjs, line 40: Are we sure this type is correct?

src/lib/server.ts
```diff
@@ -27,38 +28,65 @@ import { IncomingMessageForServer } from 'http';
...
+    body: Uint8Array;
+    token: Opt<Token>;
+};
 
-export const Callback = Func([text], StreamingCallbackHttpResponse, 'query');
+function Callback<Token extends CandidType>(token: Token) {
+    return Func([token], Opt(StreamingCallbackHttpResponse(token)), 'query');
+}
+type Callback = Func;
```
lastmjs, line 48: This type doesn't seem correct

src/lib/server.ts
```diff
@@ -27,38 +28,65 @@ import { IncomingMessageForServer } from 'http';
...
+function StreamingStrategy<Token extends CandidType>(token: Token) {
+    return Variant({
+        Callback: CallbackStrategy(token)
+    });
+}
 
-export const StreamingStrategy = Variant({
-    Callback: CallbackStrategy
-});
+export type StreamingStrategy<Token> = Variant<{
```
lastmjs, line 67: Type?

src/lib/server.ts
```diff
@@ -27,38 +28,65 @@ import { IncomingMessageForServer } from 'http';
...
+export function HttpResponse<Token extends CandidType>(token?: Token) {
+    return Record({
+        status_code: nat16,
+        headers: Vec(HeaderField),
+        body: blob,
+        streaming_strategy: Opt(StreamingStrategy(token ?? DefaultToken)),
+        upgrade: Opt(bool)
+    });
+}
+export type HttpResponse<Token> = {
```
lastmjs, line 83: Type?

src/lib/server.ts
```diff
@@ -67,6 +95,7 @@ export const HttpRequest = Record({
     body: blob,
     certificate_version: Opt(nat16)
 });
+export type HttpRequest = typeof HttpRequest.tsType;
```
lastmjs, line 98: Why can't we use `.tsType` everywhere else?

src/lib/server.ts
lastmjs, line 1: I would like to find out why the tests for http_counter were passing just fine without these changes, why was the streaming strategy working?

src/lib/server.ts
bdemann: I believe it's because the other ones have the generic Token

src/lib/server.ts
bdemann: It wasn't passing fine before these changes, part of this pr is making sure http_counter is working
https://github.com/demergent-labs/azle/pull/1533/files#diff-524c3fe13d66c09dab711f3d591a61b8e83c68eece90cffb8620193118f8a692
All of the tests were being skipped

lastmjs (approved):

src/lib/server.ts
```diff
@@ -27,38 +28,65 @@ import { IncomingMessageForServer } from 'http';
...
+        token: Opt(token)
+    });
+}
+type StreamingCallbackHttpResponse<Token> = {
+    body: Uint8Array;
+    token: Opt<Token>;
+};
 
-export const Callback = Func([text], StreamingCallbackHttpResponse, 'query');
+function Callback<Token extends CandidType>(token: Token) {
```
lastmjs, line 45: This and its type should be exported

src/lib/server.ts
lastmjs, line 40: This should be exported

lastmjs (changes requested):

lastmjs (approved):

PR #1589, opened 2024-01-26
Fs tests

lastmjs:


examples/apollo_server/src/index.ts
```diff
@@ -0,0 +1,102 @@
...
+        id: '1',
+        title: 'City of Glass',
+        rating: 63
+    }
+];
+
+const authors = [
+    {
+        id: '0',
+        name: 'Jordan'
```
bdemann, line 56: What if we updated these authors to be the authors of the above books, or updated the titles to be the titles of works created by Jordan or Ben respectively, I don't know how much I shifting the credit to authorship of those books

examples/apollo_server/test/tests.ts
```diff
@@ -0,0 +1,74 @@
+import * as dns from 'node:dns';
+dns.setDefaultResultOrder('ipv4first');
+
+import { Test } from 'azle/test';
+
+export function getTests(canisterId: string): Test[] {
+    const origin = `http://${canisterId}.localhost:8000`;
+
+    return [
+        {
```
bdemann, line 10: It troubles me a little bit that there are only two tests. I understand that we don't have time to test every facet of every api, but this is two tests testing essentially the same functionality just on two different lists

bdemann (approved):

PR #1601, opened 2024-02-01
add fetch to management

bdemann:


examples/management_canister/src/index.ts
```diff
@@ -199,3 +355,39 @@ export default Canister({
         return state.createdCanisterId;
     })
 });
+
+async function callManagementCanisterasdf(
```
lastmjs, line 359: Remove this function

examples/management_canister/src/index.ts
```diff
@@ -199,3 +355,39 @@ export default Canister({
...
+        return await response.json();
+    } else {
+        return await ic.call(managementCanister.create_canister, {
+            args: [{ settings: None, sender_canister_version: None }],
+            cycles: 50_000_000_000_000n
+        });
+    }
+}
+
+function convertOpt<T>(opt: Opt<T>): [T] | [] {
```
lastmjs, line 387: Can you make sure this is the exact same code and name as the other place this is used?

lastmjs (changes requested):

PR #1693, opened 2024-03-07
Stable infinite files

bdemann:
It takes about 15 seconds to hash a 1 GB file under the current system


examples/large_files/.gitignore
```diff
@@ -0,0 +1,5 @@
+.azle
+.dfx
+node_modules
+assets/curled
```
bdemann, line 4: get rid of that

.github/workflows/test.yml
```diff
@@ -269,6 +270,12 @@ jobs:
               shell: bash -l {0}
               working-directory: ${{ matrix.example_directories }}
               run: npm link azle
+            - name: Check Directory Permissions (Optional)
```
bdemann, line 273: get rid of this if the chmod goes well

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
```diff
@@ -0,0 +1,213 @@
+pub mod temp_chunks;
+
+use quote::quote;
+
+pub fn get_upload_assets() -> proc_macro2::TokenStream {
+    let temp_chunks_utils = temp_chunks::get_temp_chunk_utils();
```
bdemann, line 6: Do we like this approach? If so should we make it like this for the hashing as well?

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+                // The request is from an earlier upload attempt. Disregard
+                false
+            } else {
+                // The request is from the current upload attempt.
+                true
+            }
+        })
+    }
+
+    fn debug(message: &str) {
```
bdemann, line 206: should we get rid of these even if we aren't quite ready to finalize this?

bdemann:
-[ ] delete temp chunks as they are written to the final

.github/workflows/test.yml
lastmjs: Get rid of all of this

examples/large_files/test/test.ts
```diff
@@ -0,0 +1,6 @@
+import { getCanisterId, runTests } from 'azle/test';
+import { getTests } from './tests';
+
+const canisterId = getCanisterId('large_files');
```
lastmjs, line 4: Rename canister to backend

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
+import { execSync } from 'child_process';
+import { existsSync } from 'fs';
+
+async function pretest() {
+    // Edge Cases
+    // generateFileOfSize(0, 'B');
```
lastmjs, line 6: TODO explain the problem a bit

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
...
+pretest();
+
+type Unit = 'B' | 'KiB' | 'MiB' | 'GiB';
+
+function generateFileOfSize(size: number, unit: Unit) {
+    const fileName = `assets/auto/test${size}${unit}`;
+    const fileSize = `${size}${getUnitAbbreviation(unit)}`;
+    if (!existsSync(fileName)) {
+        execSync(
+            `ts-node ../../scripts/file_generator.js ${fileName} ${fileSize}`,
```
lastmjs, line 45: Let's just import this functionality and not boot up a new process

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
+import { execSync } from 'child_process';
+import { existsSync } from 'fs';
+
+async function pretest() {
+    // Edge Cases
+    // generateFileOfSize(0, 'B');
+    generateFileOfSize(1, 'B');
+    generateFileOfSize(120 * 1024 * 1024 + 1, 'B'); //One more byte than can be processed in a single hash_file_by_parts call
```
lastmjs, line 8: Update these numbers

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
...
+    generateFileOfSize(150 * 1024 * 1024 + 1, 'B'); //One more byte than can be processed in a single write_file_by_parts call
+    generateFileOfSize(2_000_001, 'B'); // One more byte that the high water mark of the readstream
```
lastmjs, line 10: Comment is out of date

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    // Weird Cases
+    // generateFileOfSize(2_000_000 * 18, 'B'); //Weird writing bound
```
lastmjs, line 13: TODOs to distinguish from regular comments

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,66 @@
...
+    // generateFileOfSize(2_000_000 * 18 + 1, 'B'); //Weird writing bound
+
+    // General Cases
+    generateFileOfSize(1, 'KiB');
```
lastmjs, line 17: TODO find out the actual maximum amount of heap we can use

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+import { Test, getCanisterId } from 'azle/test';
+import { Actor, ActorSubclass, HttpAgent } from '@dfinity/agent';
+import { hashFile } from 'azle/scripts/hash_file';
+
+export function getTests(canisterId: string): Test[] {
+    const origin = `http://${canisterId}.localhost:8000`;
+
+    return [
+        {
+            name: 'wait for things to finish uploading',
```
lastmjs, line 13: Remove the need to wait here by polling in the post_install (really just in the npx azle upload-assets comand), every few seconds or so after all requests are sent, ask the canister if it's done, only stop processing when it's done

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+        generateTest('test250MiB', 'auto', origin),
+        { ...generateTest('test500MiB', 'auto', origin), skip: true }, // We currently run out of memory with this file
+        { ...generateTest('test1GiB', 'auto', origin), skip: true }
+        // generateTest(`test${2_000_000 * 18}B`, 'auto', origin),
+        // generateTest(`test${2_000_000 * 18 + 1}B`, 'auto', origin)
+    ];
+}
+
+function generateTest(
+    fileName: string,
```
lastmjs, line 67: Shouldn't this be called filePath?

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+    nodeDir: string,
+    origin: string,
+    nodeName?: string
+): Test {
+    return {
+        name: `upload: ${fileName}`,
+        test: async () => {
+            const canisterFilePath = `assets/${fileName}`;
```
lastmjs, line 75: When constructing or composing paths, please use join from path

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+        { ...generateTest('test500MiB', 'auto', origin), skip: true }, // We currently run out of memory with this file
+        { ...generateTest('test1GiB', 'auto', origin), skip: true }
+        // generateTest(`test${2_000_000 * 18}B`, 'auto', origin),
+        // generateTest(`test${2_000_000 * 18 + 1}B`, 'auto', origin)
+    ];
+}
+
+function generateTest(
+    fileName: string,
+    nodeDir: string,
```
lastmjs, line 68: Change references of node to local

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+    origin: string,
+    nodeName?: string
+): Test {
+    return {
+        name: `upload: ${fileName}`,
+        test: async () => {
+            const canisterFilePath = `assets/${fileName}`;
+            const dir = nodeDir === '' ? '' : `/${nodeDir}`;
```
lastmjs, line 76: I assume and/or hope that join would obviate the need for this line

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+            const nodeFileName = nodeName ?? fileName;
+            const nodeFilePath = `assets${dir}/${nodeFileName}`;
+            const actor = await createGetFileHashActor(
+                getCanisterId('large_files'),
+                '8000'
+            );
+
+            const expectedHash = (await hashFile(nodeFilePath)).toString('hex');
+
+            let response = await fetch(
```
lastmjs, line 86: Why let?

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+            );
+            let exits = await response.json();
+
+            if (exits === false) {
+                return {
+                    Err: `File ${canisterFilePath} failed to upload`
+                };
+            }
+
+            const hash = await actor.get_file_hash(`${canisterFilePath}`);
```
lastmjs, line 97: We need to think about authentication tests and mainnet tests...

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+    return {
+        name: `upload: ${fileName}`,
+        test: async () => {
+            const canisterFilePath = `assets/${fileName}`;
+            const dir = nodeDir === '' ? '' : `/${nodeDir}`;
+            const nodeFileName = nodeName ?? fileName;
+            const nodeFilePath = `assets${dir}/${nodeFileName}`;
+            const actor = await createGetFileHashActor(
+                getCanisterId('large_files'),
+                '8000'
```
lastmjs, line 81: get the port number through automatic means...maybe just make a get agent host function

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+                return {
+                    Err: `File ${canisterFilePath} failed to upload`
+                };
+            }
+
+            const hash = await actor.get_file_hash(`${canisterFilePath}`);
+            return { Ok: hash === expectedHash };
+        }
+    };
+}
```
lastmjs, line 101: Please add tests for the standalone npx azle upload-assets command

examples/large_files/dfx.json
```diff
@@ -0,0 +1,29 @@
...
+        "large_files": {
+            "type": "custom",
+            "main": "src/index.ts",
+            "candid": "src/index.did",
+            "candid_gen": "http",
+            "build": "npx azle large_files",
+            "post_install": "npx azle upload-assets large_files",
+            "wasm": ".azle/large_files/large_files.wasm",
+            "gzip": true,
+            "assets_large": [
```
lastmjs, line 12: Make an issue discussing if we should combine assets and assets_large

examples/large_files/assets/permanent/text/thing.txt
lastmjs, line 1: Remove irony from text files

examples/large_files/src/index.ts
```diff
@@ -0,0 +1,32 @@
+import { Server } from 'azle';
+import { statSync, existsSync } from 'fs';
+import express, { Request } from 'express';
```
lastmjs, line 3: alphabateize

lastmjs (changes requested):

scripts/file_generator.js
lastmjs, line 1: I feel like this file could be made much simpler and more declarative. It should be rewritten in TypeScript as well. Using just appendfile or other more abstracted methods would be nice. I'm not sure there's a need to create streams and listen to events and such. Great simplification possibly here please

lastmjs: Great simplification I think is possible in the algorithm too, and let's get rid of the vars and mutations and such, I think we could make this dead simple, is my feeling

scripts/hash_file.ts
lastmjs, line 1: Same thing here, mutations, read, close, etc I have a feeling we can just use higher-level abstractions and make this simpler, for example look at the fs example for reading parts of a file

scripts/read_file.js
lastmjs, line 1: What is this script for? Do we need it? We probably should almost never be using these callback methods as we can just use the promise APIs and async/await for most operations

src/compiler/handle_cli.ts
```diff
@@ -90,3 +97,10 @@ function handleCommandClean(
 
     console.info(`azle images removed`);
 }
+
+async function handleUploadAssets() {
+    const canisterId = process.argv[3];
```
lastmjs, line 102: Isn't this the canister name and not the id?

src/compiler/rust/canister/Cargo.toml
```diff
@@ -18,11 +18,10 @@ canister_methods = { path = "../canister_methods" }
 include_dir = "0.7.3"
 slotmap = "=1.0.6"
 wasmi = "0.31.2"
+sha2 = "0.10.8"
 
 # TODO transient feature can be removed once https://github.com/wasm-forge/ic-wasi-polyfill/issues/14
-ic-wasi-polyfill = { git = "https://github.com/wasm-forge/ic-wasi-polyfill", rev = "2d2edb382816e12da9bc81b786b7cd1a00d36735", features = [
-    "transient",
-] }
+ic-wasi-polyfill = { git = "https://github.com/wasm-forge/ic-wasi-polyfill", rev = "2d2edb382816e12da9bc81b786b7cd1a00d36735" }
```
lastmjs, line 24: We want to be transient again

lastmjs (changes requested):

src/compiler/asset_uploader/index.ts
```diff
@@ -0,0 +1,210 @@
...
+            );
+        }
+
+        await throttle();
+        // Don't await here! Awaiting the agent will result in about a 4x increase in upload time.
+        // The above throttling is sufficient to manage the speed of uploads
+        actor
+            .upload_asset(destPath, timestamp, chunkNumber, chunk, size)
+            .catch((error) => {
+                if (process.env.AZLE_VERBOSE === 'true') {
```
lastmjs, line 131: Make not a verbose

src/compiler/rust/canister/src/lib.rs
```diff
@@ -74,6 +81,18 @@ thread_local! {
     static RELOADED_JS_TIMESTAMP: RefCell<u64> = RefCell::new(0);
 
     static RELOADED_JS: RefCell<BTreeMap<u64, Vec<u8>>> = RefCell::new(BTreeMap::new());
+
+    static FILE_INFO: RefCell<BTreeMap<String, (ChunkCount, BytesReceived)>> = RefCell::new(BTreeMap::new());
+
+    static UPLOADED_ASSETS_TIMESTAMP: RefCell<BTreeMap<String, u64>> = RefCell::new(BTreeMap::new());
```
lastmjs, line 87: Rename to TIMESTAMPS

src/compiler/rust/canister/src/lib.rs
```diff
@@ -74,6 +81,18 @@ thread_local! {
...
+
+    static PARTIAL_ASSETS_HASHES: RefCell<HashMap<String, Hash>> = RefCell::new(HashMap::new());
+
+    static ASSETS_HASHES: RefCell<HashMap<String, Hash>> = RefCell::new(HashMap::new());
+
+    // TODO get rid of this as soon as we have stable chunks working
+    static UPLOADED_ASSETS: RefCell<BTreeMap<String, BTreeMap<u64, Bytes>>> =
```
lastmjs, line 94: Get rid of this as part of the PR, use the filesystem

src/compiler/rust/canister/src/lib.rs
lastmjs, line 1: We must look into gzipping

lastmjs: Let's at least make an issue to look into this to potentially cut the upload time significantly

src/compiler/rust/canister_methods/src/hash_file.rs
lastmjs, line 1: Make all of this idempotent. We need to be able to engage the upload assets process over and over and over and over and over without wiping the heap. It should work each time

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,107 @@
...
+        pub fn get_file_hash(path: String) -> String {
+            ASSETS_HASHES
+                .with(|asset_hashes| asset_hashes.borrow().get(&path).unwrap().clone())
+                .iter()
+                .map(|bytes| format!("{:02x}", bytes))
+                .collect()
+        }
+
+        #[ic_cdk_macros::query]
+        pub fn get_file_hashes() -> Vec<String> {
```
lastmjs, line 20: Put in a TODO to get rid of this after testing is done

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub fn get_hash_file() -> proc_macro2::TokenStream {
+    quote! {
+        pub fn hash_file(path: String) {
+            ic_cdk::println!("START HASH FILE");
+            hash_file_by_parts(path, 0)
+        }
+
+        #[ic_cdk_macros::query]
+        pub fn get_file_hash(path: String) -> String {
```
lastmjs, line 11: Unify asset and file names

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        fn hash_file_by_parts(path: String, position: u64) {
+            ic_cdk::println!(
+                "Hash {} starting at: {}",
+                path,
+                bytes_to_human_readable(position)
+            );
+            let mut file = std::fs::File::open(&path).unwrap();
+
+            std::io::Seek::seek(&mut file, std::io::SeekFrom::Start(position)).unwrap();
```
lastmjs, line 45: I wonder if there's a better way to do this

lastmjs (changes requested):

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        let uploaded_asset_len = match write_temp_chunk(&dest_path, asset_bytes, chunk_number) {
+            Ok(len) => len as u64,
+            Err(err) => {
+                ic_cdk::println!("Error writing temp chunk: {}", err);
+                panic!("{}", err)
+            }
+        };
+
+        match test_read_temp_chunk(&dest_path, chunk_number, false) {
```
lastmjs, line 33: Why are we doing this here?

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+                dest_path
+            );
+            let delay = core::time::Duration::new(0, 0);
+            ic_cdk::println!("Spawning writer for {}", dest_path);
+            let write_closure = move || write_file_by_parts(dest_path);
+            ic_cdk_timers::set_timer(delay, write_closure);
+        }
+    }
+
+    fn write_file_by_parts(dest_path: String) {
```
lastmjs, line 58: Can we split up this algorithm into multiple declarative files

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
lastmjs, line 1: Let's make this thing more declarative

lastmjs (changes requested):

examples/large_files/dfx.json
bdemann: https://github.com/demergent-labs/azle/issues/1702

scripts/hash_file.ts
bdemann: The fs example has
- read-file-sync with readFileSync (no good, can't read large files)
- read-file with readFile (no good, same reason)
- write-file-sync (doesn't read)
- write-file (doesn't read)
- unlink-sync (doesn't read)
- unlink (doesn't read)
- rmdir-sync (doesn't read)
- rmdir (doesn't read)
- mkdir-sync (doesn't read)
- mkdir (doesn't read)
- exits-sync (doesn't read)
- create-read-stream-large-than-high-water-mark-chunked (no good, read streams doesn't guarantee chunk size)
-  create-read-stream-large-than-high-water-mark (no good, same reason)
- high-water-mark-boundary (no good, same reason)
By using a combination of start, end, and highWaterMark I could have a very good chance of getting exactly the bytes that I want and I could wait until the stream has read all of the bytes before proceeding... but I feel like would complicate it quite a bit, while this is pretty simple, open the file read exactly every time n bytes into the buffer and close the file. I can't imagine something with a read stream being any simpler

src/compiler/asset_uploader/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+}
+
+async function throttle() {
+    // We can only process about 4Mib per second. So if chunks are about
+    // 2 MiB or less then we can only send off two per second.
+    await new Promise((resolve) => setTimeout(resolve, 500)); // Should be 500 (ie 1 every 1/2 second or 2 every second)
+}
+
+function getAssetsToUpload(
```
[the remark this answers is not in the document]
bdemann: The caller of this is the CLI command. The post install script calls the CLI command and doesn't pass in arguments. Is there a different way the post install script is supposed to be calling?

src/compiler/rust/canister/src/lib.rs
bdemann: https://github.com/demergent-labs/azle/issues/1703

src/compiler/rust/canister_methods/src/upload_assets/mod.rs
bdemann: added a TODO to remove after testing

src/compiler/asset_uploader/index.ts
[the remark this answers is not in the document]
bdemann: Apparently not, we decided to just move it into handle_cli

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: I can't find one. Mostly this looks bad because I wanted to qualify the methods instead of importing them

examples/audio_and_video/.gitignore
```diff
@@ -1,3 +1,6 @@
 .azle
 .dfx
 node_modules
+src/backend/media/jordan_video_streaming_demo.mp4
```
bdemann, line 4: Why is this here?

bdemann: This is from the commits that Jordan merged into this branch. No worries

scripts/file_generator/index.ts
```diff
@@ -0,0 +1,57 @@
+import { writeFile } from 'fs-extra';
```
lastmjs, line 1: Do not use fs-extra

scripts/file_generator/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    const totalChunks = Math.ceil(sizeInBytes / defaultChunkSize);
+    for (let i = 0; i < totalChunks; i++) {
+        const remainingBytes = sizeInBytes - i * defaultChunkSize;
+        const chunkSize = Math.min(remainingBytes, defaultChunkSize);
+        const randomBytes = await createRandomBytes(chunkSize, buffer);
+        await appendFile(path, randomBytes);
+    }
+}
+
+async function createRandomBytes(
```
lastmjs, line 49: Do not mutate

scripts/file_generator/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    size: number,
+    buffer: Buffer
+): Promise<Buffer> {
```
lastmjs, line 52: Unnecessary promise

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,153 @@
...
+
+async function createGetFileHashActor(
```
lastmjs, line 131: We need to add authentication to the hashing endpoint

scripts/hash_file/index.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+async function hashFileByParts(
+    path: string,
+    position: number,
+    previousHash?: Buffer
+): Promise<Buffer> {
+    const { buffer, bytesRead } = await getBytesToHash(path, position);
+
+    if (bytesRead != 0) {
```
lastmjs, line 15: triple equals

scripts/hash_file/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    path: string,
+    position: number
+): Promise<FileReadResult<Buffer>> {
+    const file = await open(path, 'r');
+
+    // Read the bytes
+    // TODO Before having the stable file storage hooked up 120 worked. For right now 60 seems to be working. I think we could do more but I want to get everything in place before spending a lot of time fine tuning it
+    // TODO it would be great to get the size of the chunks from the canister, then we wouldn't have to every update this
+    const limit = 60 * 1024 * 1024; // Must be the same as on the canister end or hashes will not match
+    const buffer = Buffer.alloc(limit); // Allocate a Buffer for reading
```
lastmjs, line 40: check if this is mutated and change to let if it is

scripts/hash_file/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const fileReadResult = await file.read(buffer, 0, limit, position);
+    file.close();
+    return fileReadResult;
+}
+
+function hashChunkWith(data: Buffer, previous_hash?: Buffer): Buffer {
```
lastmjs, line 47: use camelcase

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
+import { Actor, ActorSubclass, HttpAgent } from '@dfinity/agent';
+import { Secp256k1KeyIdentity } from '@dfinity/identity-secp256k1';
+import { existsSync, createReadStream } from 'fs';
+import { readFile } from 'fs/promises';
+import {
+    getCanisterId,
+    getIdentityName,
+    getWebServerPort
+} from '../../../test';
```
lastmjs, line 9: Make an issue to move the location of these functions

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+type Dest = string;
+
+export async function uploadFiles(canisterName: string, paths: [Src, Dest][]) {
+    const actor = await createUploadFileChunkActor(
+        getCanisterId(canisterName),
+        getWebServerPort(),
+        getIdentityName()
+    );
+
+    const chunkSize = 2_000_000; // The current message limit is about 2 MiB
```
lastmjs, line 24: Change to MB

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+async function uploadFile(
+    srcPath: Src,
+    destPath: Dest,
+    chunkSize: number,
+    actor: ActorSubclass
+) {
+    console.info(`uploadFile: Uploading ${srcPath} to ${destPath}`);
```
lastmjs, line 77: I want to see what these logs look like

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    const uploadStartTime = process.hrtime.bigint();
+    const fileSize = (await stat(srcPath)).size;
+    for (let i = 0; i < fileSize; i += chunkSize) {
+        const fileStream = createReadStream(srcPath, {
```
lastmjs, line 81: Use file.read instead

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+            // The above throttling is sufficient to manage the speed of uploads
+            actor
+                .upload_file_chunk(destPath, uploadStartTime, i, data, fileSize)
+                .catch((error) => {
+                    console.error(error);
+                });
+            startIndex += data.length;
+        }
+    }
+    console.info(`uploadFile: finished ${srcPath}`);
```
lastmjs, line 105: add a line break

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function bytesToHumanReadable(sizeInBytes: number): string {
```
lastmjs, line 108: Redo this with reduce to get rid of the mutation

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        size /= 1024.0;
+    }
+
+    return `${size.toFixed(2)} ${suffixes[suffixes.length - 1]}`;
+}
+
+async function throttle() {
+    // We can only process about 4Mib per second. So if chunks are about
+    // 2 MiB or less then we can only send off two per second.
+    if (process.env.DFX_NETWORK === 'ic') {
```
lastmjs, line 127: This must have an if else

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        await new Promise((resolve) => setTimeout(resolve, 2_000)); // Mainnet requires more throttling. We found 2_000 by trial and error
+    }
+    await new Promise((resolve) => setTimeout(resolve, 500)); // Should be 500 (ie 1 every 1/2 second or 2 every second)
+}
+
+async function getIdentity(
```
lastmjs, line 133: We need to look into what other kinds of identities are possible, for example my demergent-labs-lastmjs-0 identity does not work

src/compiler/handle_cli.ts
```diff
@@ -97,6 +105,32 @@ function handleCommandClean(
...
+    if (srcPath !== undefined && destPath !== undefined) {
+        return [[srcPath, destPath]];
+    }
+    if (srcPath === undefined && destPath === undefined) {
+        // If both paths are undefined, look at the dfx.json for the assets to upload
+        const dfxJson = unwrap(getCanisterConfig(canisterName));
+        return dfxJson.assets_large ?? [];
+    }
+    throw new Error(
+        'The source path and destination path but be either both defined or both undefined'
```
lastmjs, line 130: Typo

test/index.ts
```diff
@@ -267,3 +273,11 @@ export function getCanisterOrigin(canisterName: string): string {
         canisterName
     )}.localhost:${getWebServerPort()}`;
 }
+
+export function getAgentHost(): string {
```
lastmjs, line 277: This should handle mainnet

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    );
+    return Secp256k1KeyIdentity.fromPem(await readFile(identityPath, 'utf-8'));
+}
+
+async function createUploadFileChunkActor(
+    canisterId: string,
+    replicaWebServerPort: string,
+    identityName: string
+): Promise<ActorSubclass> {
+    const host =
```
lastmjs, line 152: Just call getAgentHost, move the mainnet stuff into getAgentHost

lastmjs (changes requested):

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,71 @@
+import { Unit, createFileOfSize, toBytes } from 'azle/scripts/file_generator';
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/test.yml
```diff
@@ -111,6 +111,7 @@ jobs:
                     "examples/internet_identity",
                     "examples/key_value_store",
                     "examples/ledger_canister",
+                    "examples/large_files",
```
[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,64 @@
...
+    unit: Unit,
+    dir: string = 'auto'
+) {
+    const autoDir = join('assets', dir);
+    const path = join(autoDir, `test${size}${unit}`);
+    const fileSize = toBytes(size, unit);
+    await createFileOfSize(path, fileSize);
+}
+
+async function clearDir(dirPath: string, recursive: boolean = false) {
```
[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,153 @@
...
+        generateTest(origin, 'test1MiB', 'auto'),
+        generateTest(origin, 'test10MiB', 'auto'),
+        generateTest(origin, 'test100MiB', 'auto'),
+        generateTest(origin, 'test250MiB', 'auto'),
+        generateTest(origin, 'test1GiB', 'auto'),
+        // Manual Upload
+        {
+            name: 'test manual upload',
+            test: async () => {
+                await uploadFiles('backend', [
```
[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,153 @@
...
+            );
+
+            const expectedHash = (await hashFile(localFilePath)).toString(
+                'hex'
+            );
+
+            const response = await fetch(
+                `${origin}/exists?path=${canisterFilePath}`
+            );
+            const exits = await response.json();
```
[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/tests.ts
```diff
@@ -0,0 +1,153 @@
...
+
+            if (exits === false) {
+                return {
+                    Err: `File ${canisterFilePath} failed to upload`
+                };
+            }
+
+            const hash = await actor.get_file_hash(`${canisterFilePath}`);
```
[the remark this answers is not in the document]
bdemann: done

scripts/file_generator/index.ts
bdemann: done

scripts/file_generator/index.ts
bdemann: done

scripts/file_generator/index.ts
bdemann: done

examples/large_files/test/tests.ts
bdemann: done

scripts/hash_file/index.ts
bdemann: done

scripts/hash_file/index.ts
bdemann: done

src/compiler/file_uploader/index.ts
bdemann: I went ahead and just moved this

src/compiler/file_uploader/index.ts
bdemann: done

src/compiler/file_uploader/index.ts
bdemann: done

src/compiler/handle_cli.ts
bdemann: done

test/index.ts
bdemann: done

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,131 @@
...
+    quote! {
+        pub fn hash_file(path: String) {
+            ic_cdk::println!("START HASH FILE");
+            clear_partial_hash(&path);
+            clear_file_hash(&path);
+            hash_file_by_parts(path, 0)
+        }
+
+        #[ic_cdk_macros::query]
+        pub fn get_file_hash(path: String) -> String {
```
lastmjs, line 13: Let's use an Opt return type and not explode if the hash doesn't exist

src/compiler/file_uploader/index.ts
bdemann: done

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,131 @@
...
+        fn hash_chunk_with(data: &[u8], previous_hash: Option<&Vec<u8>>) -> Vec<u8> {
+            let mut h = <sha2::Sha256 as sha2::Digest>::new();
+            sha2::Digest::update(&mut h, data);
+            if let Some(hash) = previous_hash {
+                sha2::Digest::update(&mut h, hash);
+            }
+            sha2::Digest::finalize(h).to_vec()
+        }
+
+        fn load_hashes() -> Result<HashMap<String, Vec<u8>>, std::io::Error> {
```
lastmjs, line 100: I think reading from the file can be made much simpler by using simpler std read functions

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,131 @@
...
+            // Check if the file exists
+            if !std::path::Path::new(FILE_HASH_PATH).exists() {
+                return Ok(file_hashes);
+            }
+            let mut file = std::fs::File::open(FILE_HASH_PATH)?;
+
+            let mut data = Vec::new();
+            std::io::Read::read_to_end(&mut file, &mut data)?;
+
+            if !data.is_empty() {
```
lastmjs, line 111: Remove mutations

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,131 @@
...
+                file_hashes = serde_json::from_slice(&data)?; // (Example using serde)
+            }
+
+            Ok(file_hashes)
+        }
+
+        fn save_hashes(file_hashes: &HashMap<String, Vec<u8>>) -> Result<(), std::io::Error> {
```
lastmjs, line 118: Look into simpler ways to read and write

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        fn set_file_hash(path: &String, hash: Vec<u8>) {
+            let mut file_hashes = load_hashes().unwrap();
+            file_hashes.insert(path.clone(), hash);
+            save_hashes(&file_hashes).unwrap();
+        }
+
+        fn hash_chunk_with(data: &[u8], previous_hash: Option<&Vec<u8>>) -> Vec<u8> {
+            let mut h = <sha2::Sha256 as sha2::Digest>::new();
```
lastmjs, line 92: I think we can simplify the hashing, let mut h = Sha256::new()

h.update()

src/compiler/rust/canister_methods/src/hash_file.rs
lastmjs, line 1: Think about clearing all global variables and stuff written to disk

lastmjs (changes requested):

src/compiler/file_uploader/index.ts
bdemann: done

.github/workflows/test.yml
lastmjs, line 1: This is just a general comment:

```
wait for things to finish uploading

waiting 1000 milliseconds
done waiting


Running test: upload: photos/people/george-washington.tif

 test upload: photos/people/george-washington.tif failed Error: Call was rejected:
  Request ID: 0b62ef1a552b0d4906e1f25161b6db377f6e7f027df73d2217fcdd5342accffd
  Reject code: 5
  Reject text: Canister bkyz2-fmaaa-aaaaa-qaaaq-cai trapped explicitly: Panicked at 'called `Option::unwrap()` on a `None` value', canister/src/lib.rs:92:1
```

Tests are not passing for me locally, that's an error I got. I also have warnings in the replica logs

lastmjs (changes requested):

src/compiler/file_uploader/index.ts
bdemann: Although I will standby what I said about this not being as simple as the other one

src/compiler/file_uploader/index.ts
bdemann: done

src/compiler/file_uploader/index.ts
bdemann: like what?

src/compiler/file_uploader/index.ts
bdemann: done

scripts/hash_file/index.ts
bdemann: done

src/compiler/rust/canister/src/lib.rs
```diff
@@ -62,6 +62,12 @@ impl Storable for AzleStableBTreeMapValue {
     const BOUND: Bound = Bound::Unbounded;
 }
 
+type Hash = Vec<u8>;
+
+type BytesReceived = u64;
+
+const FILE_HASH_PATH: &str = "file_hashes.json";
```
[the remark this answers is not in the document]
bdemann: optionally we could also try and mimic a standard os and do something like /config/.azle/file_hashes.json? thougths?

src/compiler/rust/canister/src/lib.rs
```diff
@@ -74,6 +80,12 @@ thread_local! {
     static RELOADED_JS_TIMESTAMP: RefCell<u64> = RefCell::new(0);
 
     static RELOADED_JS: RefCell<BTreeMap<u64, Vec<u8>>> = RefCell::new(BTreeMap::new());
+
+    static FILE_INFO: RefCell<BTreeMap<String, BytesReceived>> = RefCell::new(BTreeMap::new());
```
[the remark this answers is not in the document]
bdemann: done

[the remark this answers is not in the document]
bdemann: Everything in FILE_INFO gets set to default values at the start of a new upload and completely deleted at the end of a hash.

Done

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+use crate::hash_file;
+
+pub fn get_upload_file_chunk() -> proc_macro2::TokenStream {
+    let write_chunk_src = write_chunk::get_write_chunk();
+    let verify_latest_src = verify_latest::get_verify_latest();
+    let hash_file_src = hash_file::get_hash_file();
+
+    quote! {
+        #write_chunk_src
+        #verify_latest_src
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            timestamp: u64,
+            start_index: u64,
+            file_bytes: Vec<u8>,
+            total_file_len: u64,
+        ) {
+            if !ic_cdk::api::is_controller(&ic_cdk::api::caller()) {
+                panic!("Must be a controller to upload files!");
+            }
+
+            let is_current_timestamp = verify_latest_version(&dest_path, timestamp);
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+            let uploaded_file_len = match write_chunk(&dest_path, file_bytes, start_index, total_file_len) {
+                Ok(len) => len as u64,
+                Err(err) => {
+                    ic_cdk::println!("Error writing chunk: {}", err);
+                    panic!("{}", err)
+                }
+            };
+
+            ic_cdk::println!(
```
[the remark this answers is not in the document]
bdemann: No can do, we need uploaded_file_len from the call above, I updated the log to not be as confusing

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            );
+
+            if uploaded_file_len == total_file_len {
+                ic_cdk::println!(
+                    "upload_file_chunk: UPLOAD COMPLETE for {}. Spawning hash task",
+                    dest_path
+                );
+
+                let delay = core::time::Duration::new(0, 0);
+                let hash_closure = move || hash_file(dest_path);
```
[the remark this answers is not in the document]
bdemann: no, but the other one is

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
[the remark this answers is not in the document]
bdemann: Do to the not quite synchronous nature of the canister side the line breaks don't really help... They might hinder more than anything because they group things up into things that are not actually groups. I'll leave them in and you can render an opinion during the final review

src/compiler/rust/canister_methods/src/upload_file_chunk/verify_latest.rs
```diff
@@ -0,0 +1,47 @@
...
+                    // Unless they don't get overwritten properly and the chunk-writer tries to use them to create the next version of this file.
+                    ic_cdk::println!(
+                        "WARNING: Failed to delete file {} uploaded at {}. {}",
+                        dest_path,
+                        current_timestamp,
+                        err
+                    );
+                }
+                true
+            } else if current_timestamp < last_recorded_timestamp {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/write_chunk.rs
```diff
@@ -0,0 +1,61 @@
...
+            let new_file = std::fs::OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open(&path)?;
+            new_file.set_len(file_len)?;
+            Ok(new_file)
+        }
+
+        fn update_file_info(dest_path: &str, bytes_in_chunk: usize) {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/write_chunk.rs
```diff
@@ -0,0 +1,61 @@
...
+                Err(_) => init_file(path, file_len)?,
+            };
+
+            std::io::Seek::seek(&mut file, std::io::SeekFrom::Start(start_index))?;
+            std::io::Write::write_all(&mut file, &file_bytes)?;
+            drop(file);
+
+            update_file_info(path, file_bytes.len());
+
+            Ok(get_total_bytes_written(path))
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/write_chunk.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        fn init_file(path: &str, file_len: u64) -> std::io::Result<std::fs::File> {
+            let new_file = std::fs::OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
```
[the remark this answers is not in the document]
bdemann: Nope, we need it to clear the file if it already exists.

src/compiler/rust/canister_methods/src/upload_file_chunk/verify_latest.rs
```diff
@@ -0,0 +1,47 @@
+use quote::quote;
+
+pub fn get_verify_latest() -> proc_macro2::TokenStream {
```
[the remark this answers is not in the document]
bdemann: Also don't forget to have the hash function be authenticated it needs to use the caller information we get from the agent

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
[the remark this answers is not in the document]
bdemann: it was because the file didn't exist, we do a check for that now. Now that we are writing to the file directly I think it's safe to upgrade that warning to a failure. I have done that now.

src/compiler/rust/canister_methods/src/upload_file_chunk/verify_latest.rs
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+
+        fn reset_for_new_upload(path: &str, timestamp: u64) -> std::io::Result<()> {
+            UPLOADED_FILE_TIMESTAMPS.with(|upload_file_timestamps_map| {
+                let mut upload_file_timestamps_map_mut = upload_file_timestamps_map.borrow_mut();
+
+                upload_file_timestamps_map_mut.insert(path.to_string(), timestamp);
+            });
+            // TODO check if the file exists
```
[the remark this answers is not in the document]
bdemann: it was because the file didn't exist, we do a check for that now. Now that we are writing to the file directly I think it's safe to upgrade that warning to a failure. I have done that now.

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: I got you
`let mut h: sha2::Sha256 = sha2::Digest::new();`
And h.update() doesn't exist without importing

To my knowledge the reason we are Qualifying everywhere instead of Importing is to make it easier to manage imports since things are spread out among many almost unrelated files because of using the quotes. 

If this is not the case or we actually don't think that's important anymore and we would rather Import that would clean up a lot of stuff.

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: done

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: apparently std::fs provides some convenience functions that combine std::io::open with std::io::write or std::io::read. I'll use those instead where I can

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/verify_latest.rs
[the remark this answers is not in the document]
bdemann: done on both accounts

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+                panic!("Must be a controller to upload files!");
+            }
+
+            let is_current_timestamp = verify_latest_version(&dest_path, timestamp);
+
+            if !is_current_timestamp {
+                return;
+            }
+
+            let uploaded_file_len = match write_chunk(&dest_path, file_bytes, start_index, total_file_len) {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: We're going to hook into the end of the process and say "hey canister we are done uploading you should have everything," and the canister can be like, "but there are these files that I haven't finished" and we'll be like, neat lets report that to the user. Next we will take a list of the files that uploaded successfully and make sure the hashes are done, if not set a timer to check again later, and keep doing that until the hashes are done and then we end the process.

src/compiler/rust/canister/src/lib.rs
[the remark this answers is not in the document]
bdemann: yes let's do this!

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,64 @@
...
+import { join } from 'path';
+
+async function pretest() {
+    await clearDir(join('assets', 'auto'));
+    // Edge Cases
+    // TODO excluded because it will require some reworking to get 0 byte files to work and it doesn't seem urgent
+    // generateFileOfSize(0, 'B');
+    await generateFileOfSize(1, 'B');
+    await generateFileOfSize(60 * 1024 * 1024 + 1, 'B'); //One more byte than can be processed in a single hash_file_by_parts call
+    await generateFileOfSize(2_000_001, 'B'); // One more byte that the high water mark of the readstream
```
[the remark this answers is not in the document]
bdemann: The question here is "Is the use of the term 'High Water mark' appropriate in this context?" If so leave it here, if not, then get rid of it.

Cargo.lock
```diff
@@ -225,6 +225,8 @@ dependencies = [
  "ic-stable-structures",
  "ic-wasi-polyfill",
  "include_dir",
+ "serde_json",
```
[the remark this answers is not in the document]
bdemann: we've decided to use the filesystem, that way we don't have to reserve more spots that developers might use for stable memory

src/compiler/file_uploader/index.ts
bdemann: user side "Uploading chunk:" canister sided "Received chunk:"
get rid of start and finish messages.
just hashing: asset/path | start at

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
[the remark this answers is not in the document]
bdemann: We decided that on the canister end we will get rid of the line breaks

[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/pretest.ts
```diff
@@ -0,0 +1,64 @@
...
+    // General Cases
+    // TODO Add tests for huge files after https://github.com/wasm-forge/stable-fs/issues/2 is resolved
+    await generateFileOfSize(1, 'KiB');
+    await generateFileOfSize(10, 'KiB');
+    await generateFileOfSize(100, 'KiB');
+    await generateFileOfSize(1, 'MiB');
+    await generateFileOfSize(10, 'MiB');
+    await generateFileOfSize(100, 'MiB');
+    await generateFileOfSize(250, 'MiB');
+    await generateFileOfSize(1, 'GiB');
```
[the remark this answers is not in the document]
bdemann: done

examples/large_files/test/pretest.ts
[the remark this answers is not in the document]
bdemann: done

dfx/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    const agent = new HttpAgent({
+        host: getAgentHost()
+    });
+
+    if (process.env.DFX_NETWORK !== 'ic') {
+        await agent.fetchRootKey();
+    }
+}
+
+export async function createAgent(): Promise<HttpAgent> {
```
lastmjs, line 44: createAuthenticatedAgent

examples/large_files/test/tests.ts
```diff
@@ -56,11 +58,14 @@ export function getTests(canisterId: string): Test[] {
...
-                ]);
+                execSync(
+                    `npx azle upload-files backend assets/manual/test150MiB assets/test150MiB`,
+                    {
+                        stdio: 'inherit'
+                    }
+                );
 
-                await new Promise((resolve) => setTimeout(resolve, 10 * 1000));
+                await new Promise((resolve) => setTimeout(resolve, 1 * 1000));
```
lastmjs, line 68: Remove this

examples/large_files/test/tests.ts
lastmjs: This seems important, we need to prove that bad users can't call the method

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        #[ic_cdk_macros::update]
+        pub fn upload_file_chunk(
+            dest_path: String,
+            timestamp: u64,
+            start_index: u64,
+            file_bytes: Vec<u8>,
+            total_file_len: u64,
+        ) {
+            if !ic_cdk::api::is_controller(&ic_cdk::api::caller()) {
```
lastmjs, line 26: Let's use a guard, or at least create a function that has this logic in it and use that everywhere

lastmjs (changes requested):

dfx/index.ts
bdemann: done

examples/large_files/test/tests.ts
bdemann: done

src/compiler/file_uploader/bytes_to_human_readable.ts
```diff
@@ -0,0 +1,26 @@
+export function bytesToHumanReadable(sizeInBytes: number): string {
+    const suffixes = ['B', 'KiB', 'MiB', 'GiB'];
+
+    const result = suffixes.reduce(
+        (acc, suffix) => {
+            if (acc.done) {
+                return acc;
+            }
+            if (acc.size < 1024.0) {
+                acc.unit = suffix;
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/expand_paths.ts
```diff
@@ -0,0 +1,50 @@
...
+        },
+        Promise.resolve([])
+    );
+}
+
+async function expandPath(
+    srcPath: Src,
+    destPath: Dest
+): Promise<[Src, Dest][]> {
+    if (!existsSync(srcPath)) {
```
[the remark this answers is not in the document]
bdemann: Using `fs.exists` is deprecated. Instead, the current recommendation is to
directly attempt the file operation and handle errors with a `try...catch`
block. This approach eliminates the possibility of the file's existence state
changing between checking and performing the operation. Since we're throwing an
error if the file doesn't exist, I'm going to get rid of the throw and simply
let `fs` handle it for us.

src/compiler/file_uploader/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    for (const [srcPath, destPath] of expandedPaths) {
+        // Await each upload so the canister doesn't get overwhelmed by requests
+        await uploadFile(srcPath, destPath, chunkSize, actor);
+    }
+
+    onBeforeExit(canisterId, expandedPaths);
+    onExit(canisterId, expandedPaths);
+}
+
+async function createUploadFileChunkActor(
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/on_exit.ts
```diff
@@ -0,0 +1,35 @@
...
+            const actor = await createClearFileAndInfoActor(canisterId);
+            await actor.clear_file_and_info(path);
+        }
+    });
+}
+
+async function createClearFileAndInfoActor(
+    canisterId: string
+): Promise<ActorSubclass> {
+    const agent = await createAgent();
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/handle_cli.ts
```diff
@@ -97,6 +105,32 @@ function handleCommandClean(
...
 
+async function handleUploadFiles() {
+    const canisterName = process.argv[3];
+    const srcPath = process.argv[4];
+    const destPath = process.argv[5];
+    const filesToUpload = getFilesToUpload(canisterName, srcPath, destPath);
+    await uploadFiles(canisterName, filesToUpload);
+}
+
+function getFilesToUpload(
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister/src/lib.rs
```diff
@@ -62,6 +62,12 @@ impl Storable for AzleStableBTreeMapValue {
     const BOUND: Bound = Bound::Unbounded;
 }
 
+type Hash = Option<Vec<u8>>;
+type Timestamp = u64;
+type BytesReceived = u64;
+
+const FILE_HASH_PATH: &str = ".config/azle/file_hashes.json";
```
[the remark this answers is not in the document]
bdemann: It seems like it is currently not supported to have Path::new() (which we would need to use join) with a const like this.
https://github.com/rust-lang/rust/pull/92930
https://stackoverflow.com/questions/76348617/why-is-it-not-possible-to-declare-a-static-or-constant-stdpathpath-object-in

Options:
1. Since we are currently only targeting canisters this file format should work for everything and we can just keep it like that for the time being.
2. We could just recreate this variable in each of the three places we use it.
3. We could do a hybrid where the parts of the path are strings and then at each use we make it into a path
4. We could make a function for it.
     5. I like that one, I think I'll try it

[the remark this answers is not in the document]
bdemann: Something like this perhaps?
```rust
pub fn get_file_hash_path() -> std::path::PathBuf {
    std::path::Path::new(".config")
        .join("azle")
        .join("file_hashes.json")
}
```

src/compiler/rust/canister_methods/src/upload_file_chunk/versions.rs
```diff
@@ -0,0 +1,60 @@
...
+    quote! {
+        pub fn check_if_latest_version(dest_path: &str, current_timestamp: Timestamp) -> bool {
+            let last_recorded_timestamp = get_timestamp(dest_path);
+
+            if current_timestamp < last_recorded_timestamp {
+                // The request is from an earlier upload attempt. Disregard
+                return false;
+            }
+
+            if current_timestamp > last_recorded_timestamp {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/incomplete_files.ts
```diff
@@ -0,0 +1,41 @@
...
+import { createAgent } from '../../../dfx';
+
+export async function getListOfIncompleteFiles(
+    paths: [Src, Dest][],
+    canisterId: string
+): Promise<[Src, Dest][]> {
+    const filters = await Promise.all(
+        paths.map(async ([_, destPath]) => {
+            try {
+                let hashActor = await createGetFileHashActor(canisterId);
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/incomplete_files.ts
```diff
@@ -0,0 +1,41 @@
+import { ActorSubclass, Actor } from '@dfinity/agent';
+import { Src, Dest } from '.';
+import { createAgent } from '../../../dfx';
+
+export async function getListOfIncompleteFiles(
+    paths: [Src, Dest][],
+    canisterId: string
+): Promise<[Src, Dest][]> {
+    const filters = await Promise.all(
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
bdemann: done

src/compiler/rust/canister_methods/src/upload_file_chunk/versions.rs
```diff
@@ -0,0 +1,60 @@
...
+            FILE_INFO.with(|uploaded_file_timestamps_map| {
+                match uploaded_file_timestamps_map.borrow().get(path) {
+                    Some((timestamp, _, _)) => timestamp.clone(),
+                    None => 0,
+                }
+            })
+        }
+
+        #[ic_cdk_macros::update]
+        pub fn clear_file_and_info(path: String) {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/clear_file_and_info_actor.ts
```diff
@@ -0,0 +1,20 @@
+import { Actor, ActorSubclass } from '@dfinity/agent';
+import { createAuthenticatedAgent } from '../../../dfx';
+
+export async function createClearFileAndInfoActor(
+    canisterId: string
+): Promise<ActorSubclass> {
```
lastmjs, line 6: Doesn't this return type remove useful information?

src/compiler/file_uploader/hash_status_actor.ts
```diff
@@ -0,0 +1,22 @@
+import { Actor, ActorSubclass } from '@dfinity/agent';
+import { createAuthenticatedAgent } from '../../../dfx';
+
+export async function createActor(canisterId: string): Promise<ActorSubclass> {
```
lastmjs, line 4: Doesn't this return type remove useful information?

src/compiler/file_uploader/incomplete_files.ts
```diff
@@ -1,31 +1,37 @@
...
     );
     return paths.filter((_, index) => filters[index]);
 }
 
+async function hasValidHash(
+    path: string,
+    actor: ActorSubclass
+): Promise<boolean> {
+    try {
+        const hashOption = (await actor.get_file_hash(path)) as [] | [string];
```
lastmjs, line 24: We shouldn't need to cast these types should we? The actor should be typed appropriately

src/compiler/rust/canister_methods/src/hash_file.rs
```diff
@@ -30,6 +32,7 @@ pub fn get_hash_file() -> proc_macro2::TokenStream {
                 bytes_to_human_readable(position),
                 bytes_to_human_readable(file_length),
             );
+            // TODO add percentage here
```
lastmjs, line 35: Was percentage added here?

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -50,6 +46,13 @@ pub fn get_upload_file_chunk() -> proc_macro2::TokenStream {
             }
         }
 
+        pub fn is_authenticated() -> Result<(), String> {
+            if ic_cdk::api::is_controller(&ic_cdk::api::caller()) {
+                return Ok(());
+            }
+            return Err("Access denied: Must be a controller to handle files!".to_string());
```
lastmjs, line 53: Can we make this generic? Just `Not Authorized: must be a controller to call this method`

src/compiler/rust/canister_methods/src/upload_file_chunk/mod.rs
```diff
@@ -15,18 +15,14 @@ pub fn get_upload_file_chunk() -> proc_macro2::TokenStream {
         #check_if_latest_version_src
         #hash_file_src
 
-        #[ic_cdk_macros::update]
+        #[ic_cdk_macros::update(guard = is_authenticated)]
```
lastmjs, line 18: Do you have tests for each of the functions with this guard to ensure that unauthorized users can't access it?

src/compiler/file_uploader/on_before_exit.ts
```diff
@@ -1,41 +1,81 @@
...
+            incompleteDestPaths
+        );
+
+        const wasUpdated = ongoingHashInfo.every(
+            (fileInfo) => fileInfo.triesSinceLastChange < 5
         );
-        return false;
+
+        if (!wasUpdated) {
+            // TODO get hashing percentage and report, if not update after 5 times end the process
```
lastmjs, line 68: Let's do this

src/compiler/rust/canister_methods/src/hash_file.rs
lastmjs, line 1: Once all tests are added, percentages added etc and code is done I want to review the hashing system again and try it out locally on my machine

lastmjs (changes requested):

lastmjs:
Please address items in this issue: https://github.com/demergent-labs/azle/issues/1717

src/compiler/file_uploader/incomplete_files.ts
bdemann: done

src/compiler/file_uploader/hash_status_actor.ts
bdemann: done

src/compiler/file_uploader/clear_file_and_info_actor.ts
bdemann: done

src/compiler/rust/canister_methods/src/hash_file.rs
bdemann: done

src/compiler/file_uploader/on_before_exit.ts
bdemann: It was already done, I just need to remove the TODO

examples/large_files/test/tests.ts
bdemann: done

src/compiler/file_uploader/bytes_to_human_readable.ts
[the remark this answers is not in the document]
bdemann: done

src/compiler/handle_cli.ts
```diff
@@ -31,6 +33,12 @@ export function handleCli(
         return true;
     }
 
+    if (commandName === 'upload-files') {
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/index.ts
bdemann: https://github.com/demergent-labs/azle/issues/1719 This doesn't seem critical for first launch. It will go from no users being able to upload to some users being able to upload. So it seems like, at least for organizations sake, we can separate this into a different feature

examples/large_files/test/tests.ts
```diff
@@ -2,18 +2,109 @@ import * as dns from 'node:dns';
...
+            name: 'Set up unauthorized user',
+            prep: async () => {
+                generateIdentity(unauthorizedUser);
+            }
+        },
+        {
+            name: 'Unauthorized Upload from actor',
+            test: async () => {
+                const destPath = 'assets/unauthorizedAddition';
+                let actor = await createActor(
```
lastmjs: const

lastmjs (changes requested):

examples/large_files/test/tests.ts
bdemann: done

src/compiler/file_uploader/ongoing_hashes.ts
```diff
@@ -12,20 +14,56 @@ type Total = bigint;
...
+
+    const incompleteHashingJobs = await updateOngoingHashingJobs(
+        previouslyOngoingJobs,
+        incompleteDestPaths,
+        actor
+    );
+
+    const areStillOngoing = incompleteHashingJobs.every(
+        // If the hash hasn't progressed in the last 5 tries it is considered failed instead of ongoing
+        (fileInfo) => fileInfo.triesSinceLastChange < 5
```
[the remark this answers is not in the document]
bdemann: done

src/compiler/file_uploader/uploader_actor.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+    );
+}
+
+interface _SERVICE {
+    clear_file_and_info: ActorMethod<[string], void>;
+    get_file_hash: ActorMethod<[string], [] | [string]>;
+    get_hash_status: ActorMethod<[string], [] | [[bigint, bigint]]>;
+    upload_file_chunk: ActorMethod<
+        [string, bigint, bigint, Buffer, bigint],
```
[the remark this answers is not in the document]
bdemann: done

PR #1697, opened 2024-03-08
Server refactor

lastmjs:
The main purpose of this PR is to make the server more modular and composable so that the end developer can create hybrid HTTP/Candid canisters easily. Multiple ways to do this are shown and tested.


bdemann:
# What is the feature?

## Does it have an issue?
No

## What is it doing?

>The main purpose of this PR is to make the server more modular and composable 
>so that the end developer can create hybrid HTTP/Candid canisters easily. Multiple
>ways to do this are shown and tested.

Seems straight forward enough

# Expectations
## Tests

- A canister that is a server with a custom init method
- A canister that is a server with a custom post upgrade method
- A canister that is a server with both an init and post upgrade method
- A canister that is a server with update and query methods
- A canister that is a server with update, query, init and post upgrade methods

More variations than that might be overkill. I think the above cases should be
sufficient.

I expect init, post upgrade and update methods to set some sort of state to
prove that they were called, and have some sort of way to get that out, either
by a server endpoint or a query method.

## Code
I think that is all I expect to see and I don't expect this pr to be very in
depth it seems pretty straight forward. I should see a pretty straight forward
way to add these methods that is simple for a user to understand.

The user in this case will be someone with enough IC knowledge to know that
init, post-upgrade, update, and query methods are a thing, so we can expect a
decent amount of IC knowledge when designing the API.

examples/hybrid_canister/src/canister.ts
```diff
@@ -0,0 +1,25 @@
+import { Canister, query, serverCanisterMethods, text, update } from 'azle';
+import express from 'express';
+
+export default Canister({
+    ...serverCanisterMethods(serverCallback),
+    candidQuery: query([], text, () => {
```
bdemann, line 6: For both purposes of testing and illustrating how to do it, I think it might be good to show either the init or post upgrade method being overwritten from the canister context

src/lib/server.ts
```diff
@@ -88,60 +106,71 @@ export type HttpResponse<Token> = {
...
+    {
+        manual: true
+    }
+);
+
+export function serverCanisterMethods(serverCallback: ServerCallback) {
+    return {
+        init: serverInit(serverCallback),
+        postUpgrade: serverPostUpgrade(serverCallback),
+        http_request: serverHttpRequest,
```
bdemann, line 151: Why are these snake_case but postUpgrade is camelCase?

examples/hybrid_canister/test/tests.ts
```diff
@@ -0,0 +1,155 @@
+import * as dns from 'node:dns';
+dns.setDefaultResultOrder('ipv4first');
+
+import { getCanisterId, Test } from 'azle/test';
+import { createActor as createActorServer } from './dfx_generated/server';
+import { createActor as createActorServerInitAndPostUpgrade } from './dfx_generated/server_init_and_post_upgrade';
+import { createActor as createActorCanister } from './dfx_generated/canister';
+
+export function getTests(): Test[] {
+    const canisterIdServer = getCanisterId('server');
```
bdemann, line 10: These should all be broken up into their own files canisterTests.ts serverTests.ts and serverInitAndPostUpgradeTests.ts

examples/hybrid_canister/test/tests.ts
```diff
@@ -0,0 +1,155 @@
...
+                        }
+                    );
+                    const httpUpdateResponseText =
+                        await httpUpdateResponse.text();
+
+                    const candidQueryText = await actorServer.candidQuery();
+                    const candidUpdateText = await actorServer.candidUpdate();
+
+                    return {
+                        Ok:
```
bdemann, line 63: These should be broken up into each their own tests, it will be easier to read and easier to figure out where the error is should any of the tests fail

examples/hybrid_canister/test/tests.ts
```diff
@@ -0,0 +1,155 @@
...
+                } catch (error: any) {
+                    return {
+                        Err: error
+                    };
+                }
+            }
+        },
+        {
+            name: 'server_init_and_post_upgrade',
+            test: async () => {
```
bdemann, line 78: We really ought to test the post_upgrade in this canister to make sure setNodeServer is called correctly and that when it is called from this context it behaves as we expect.

examples/hybrid_canister/test/tests.ts
bdemann, line 1: I would really like to see two more test canisters. One with just an init method, and one with just a post upgrade method. I want to make sure all possible combinations are tested so we are confident that setNodeServer is always called in all contexts. I wouldn't mind seeing one additional test with init and post_upgrade but no query or update methods, but I don't think that one is as crucial

examples/hybrid_canister/src/canister.ts
bdemann: And I would like to see that as an additional canister, ie (canister_init_and_post_upgrade.ts)

bdemann (changes requested):
# Results
## Code
Things looked pretty good, just a couple of consistency and documentation
concerns.

Is there any benefit into thinking about how I would make the API?
```javascript
export default Server(
    () => {
        const app = express();

        app.get('/get-endpoint', (req, res) => {
            res.send('Hello world');
        });

        return app.listen();
    },
    {
        init: init(
            [User, Reaction, Principal],
            (initUser, initReaction, initOwner) => {
                user = Some(initUser);
                reaction = Some(initReaction);
                owner = Some(initOwner);
            }
        ),
        pug: post_upgrade(
            [User, Reaction, Principal],
            (initUser, initReaction, initOwner) => {
                user = Some(initUser);
                reaction = Some(initReaction);
                owner = Some(initOwner);
            }
        )
        myUpdate: update([], [], () => {})
        myQuery: query([], [], () => {})
    }
);
```
I suppose I would do it like that. First with the server callback, then with a
body like Canister that would be optional. Cool, that's what it looks like has
been done!

## Tests
### Examples
Everything is in one example hybrid_canister.
Hybrid canister has according to the dfx.json 3 canisters
- server
- server_init_and_post_upgrade
- canister

#### Canister
Ah, canister is testing from the other way around. Instead of asking what if the
server could have canister methods, it's asking what if a canister could have
server endpoints.

Good idea!

That does add another layer to the tests I would hope to see to be exhaustive in
the coverage. I would at least like to see an init method used on the canister
side in addition to using serverCanisterMethods to put them in there. Especially
since it's going to be critical that developers add the setNodeServer call to
their init and post_upgrade methods.

#### Server
Server is the same as canister and is what I was expecting from the server
perspective.

#### Server Init and Post Upgrade
Server Init and Post Upgrade is Server but with a init and post upgrade method.
Looks good.

#### Missing Tests?

##### Combinations
There are a few combinations that aren't represented that I think would be good to test.

A canister that is a server with:
- [ ] a custom init method
- [ ] a custom post upgrade method
- [ ] both an init and post upgrade method
- [x] update and query methods
- [x] update, query, init and post upgrade methods

Maybe adding all three of the missing ones (just init, just post-upgrade, and
just init and post upgrade with not query or update) is a little overkill, but I
think at least one of them would be beneficial. I would probably at least one of
just init or just post upgrade as a minimum.

##### Shared Data
I think it could be interesting to test or show the different endpoint/canister
methods working on the same data and not just separate data that is only every
touched by one half of the canister.

### Test code
After looking at the test examples I can foresee that the actual tests should
be very straight forward. Each of those three examples have four endpoints or
canister methods, But the post_upgrade tests should be called twice so there
should be a total of 16 tests that should be simply calling the endpoints or
methods and checking the results. I'm not expecting anything complicated here.

I would expect more given my requests above, but given the current state of the
examples that's all I would expect.

Unsurprisingly such an approach would be a little cluttered since we are testing
three independent canisters in one test. I think it would be much easier to
read if the tests where broken up into one test file for each canister and then
each endpoint could have it's own simple test instead of being part of one big
conglomeration of all the endpoints of that canister.

The post upgrade function is not being tested at all.

Which, thinking of the ramifications of that makes me feel more strongly that
we ought to test both a server with just an init override and a server with just
a post upgrade override because I want to be sure that both of those functions
are called regardless of the combination they are in since the call to
setNodeServer is so crucial. Here we don't even deploy it again so we wouldn't
even get the error warning that nodeServer was undefined.

lastmjs:
I'd like to see this conditional receive the same declarative treatment as https://github.com/demergent-labs/azle/pull/1699, @bdemann wants this on the if statements on server just like the if statements on fetch_ic

examples/internet_identity/src/frontend/index.ts
```diff
@@ -1,6 +1,6 @@
 import { Identity } from '@dfinity/agent';
 import { AuthClient } from '@dfinity/auth-client';
-import { toJwt } from 'azle/client';
+import { toJwt } from 'azle/http_client';
```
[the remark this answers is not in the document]
lastmjs: I think we've talked about this a lot

src/lib/server.ts
```diff
@@ -25,10 +25,31 @@ import { ServerResponse, Server as NodeServer } from 'http';
 import { HttpConn } from 'http';
 // @ts-ignore
 import { IncomingMessageForServer } from 'http';
+import { CanisterOptions } from './candid/types/reference/service/canister_function';
 
 const httpMessageParser = require('http-message-parser');
 
-export const DefaultToken = Record({});
+export type HeaderField = [text, text];
```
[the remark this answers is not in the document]
lastmjs: Indeed yes, to my satisfaction that is complete

src/lib/server.ts
```diff
@@ -25,10 +25,31 @@ import { ServerResponse, Server as NodeServer } from 'http';
...
+
+export const HttpUpdateRequest = Record({
+    method: text,
+    url: text,
+    headers: Vec(HeaderField),
+    body: blob
+});
+export type HttpUpdateRequest = typeof HttpUpdateRequest.tsType;
+
+export const DefaultToken = blob;
```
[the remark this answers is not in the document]
lastmjs: I will tell you. Record means we will have to change the Candid every time we want a different token. Blob allows us to serialize anything, thus our tokens, as long as they can be serialized, wouldn't ever need to change the candid. We lose typing for some nice flexilibity, which I thought seemed like a nice trade-off

src/lib/server.ts
```diff
@@ -88,60 +106,71 @@ export type HttpResponse<Token> = {
...
+        ...serverCanisterMethods(serverCallback),
+        ...canisterOptions
     });
 }
 
 export async function httpHandler(
     httpRequest: HttpRequest | HttpUpdateRequest,
     query: boolean
 ) {
+    if (nodeServer === undefined) {
```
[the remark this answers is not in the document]
lastmjs: I haven't done any documentation yet, we will add documentation later. I will put this into my documentation issue.

src/lib/server.ts
lastmjs: Because the actual canister method names are used to find http_request and http_request_update, whereas a Rust macro is used for init and postUpgrade, so we don't need to use any specific names

examples/hybrid_canister/test/tests.ts
lastmjs: The file is very short and there is just one method to test, I think that would add unnecessary complexity to the tests

examples/hybrid_canister/test/tests.ts
lastmjs: The file is very short and there is just one method to test for each, I think that would add unnecessary complexity to the tests.

I don't see how it would be easier to figure out where the error is, the test tell you the exact test that fails, and there are only 3 tests right now

lastmjs: Ignore the previous comment, this has been addressed

examples/hybrid_canister/test/tests.ts
lastmjs: Ignore the previous comment, this has been addressed

examples/hybrid_canister/test/tests.ts
lastmjs: This is now tested

examples/hybrid_canister/src/canister.ts
lastmjs: I have implemented this additional canister: canister_init_and_post_upgrade.ts

I don't think we need to do either init or post_upgrade for now

examples/hybrid_canister/test/tests.ts
lastmjs: I don't think we need to do either init or post_upgrade for now

examples/hybrid_canister/test/tests.ts
bdemann: We didn't actually do this for the reasons above

src/lib/server.ts
```diff
@@ -320,6 +307,42 @@ export async function httpHandler(
     nodeServer.emit('request', req, res);
 }
 
+function shouldUpgrade(
+    httpRequest: HttpRequest | HttpUpdateRequest,
+    query: boolean
+): boolean {
+    const forceQueryHeaderExists = forceHeaderExists(
+        'x-ic-force-query',
```
bdemann, line 315: Can we do a toLowerCase comparison?

bdemann (changes requested):

bdemann (approved):

bdemann (approved):

Issue #1691, opened 2024-03-04
ETH LatAm


lastmjs:
- [x] Composable and abstracted server
- [x] fetch/xmlhttp authentication
- [x] for nodejs, android, etc developers can use the agent directly...make sure to export the agent and its types so that developers can use that directly if they need to
- [x] make --native azle flag for native compilation
  - [x] should we add tests for native compilation? Maybe just choose a couple examples to do this with?
- [x] Fix logging of error, error.message should log with console.log(error)
- [x] fix production canister url in terminal
- [x]  show the production url when --network is ic in the terminal
- [x] fetch http and file on backend
- [x] get rid of the need for Server
- [x] update dfx
- [x] get rid of all await new Promise() things in tests
- [x] Aggro team from ETH Latam seems to have found a bug with large files and Express static, about 600 KiB file runs out of instructions
- [x] we should log the file uploading process, we should tell the user what is happening in the terminal and in the replica
- [x] Let's add good tests for backend fetch http outcalls?
  - [x] I was using the cat APIs, there's a github repo with free APIs
  - [x] Let's make these tests pretty good and robust
  - [x] Test HEAD, GET, and POST
  - [x] Test string or URL for input
  - [x] test a bunch of little fun free APIs
  - [x] test tECDSA???
- [x] We need to discuss what to do with headers in the ethers.js fetch thing...it seems very likely that if we leave them in by default, then this will all fail in production. But I don't see a granular-enough way to deal with this right now: I have made an issue for this: https://github.com/demergent-labs/azle/issues/1716
- [x] make the ic evm rpc example really good
  - [x] static types
  - [x] update to latest did and such
- [x] Update did files for http, we now have httpupdaterequest again
- [x] create ethers signer and provider
  - [x] think more deeply about the ethers dependency and peer dependencies
  - [x] Use this to help implement and test http fetch
  - [x] create a beautiful base l2 example, purchasing things, show people how to use L2 USDC
- [x] https://github.com/demergent-labs/azle/issues/1698
- [x] Kybra
  - [x] remove custom chunk wasm uploading
  - [x] podman/docker
  - [x] assets
  - [x] large assets
- [x] large assets
- [x] ensure we can download any size file from a canister
- [x] Reload js will not work on mainnet
- [x] dfx extensions

PR #1700, opened 2024-03-09
Native compilation

lastmjs:


bdemann:
# What is the feature?

## Does it have an issue?
No. It's just a sub point of [#1691](https://github.com/demergent-labs/azle/issues/1691)

## What is it doing?
### Explanation
> - make --native azle flag for native compilation
>    - should we add tests for native compilation? Maybe just choose a couple examples to do this with?

### Questions
What on Earth is native compilation?

It looks like it is compiling the old way without podman/docker. I am guessing
we are adding this for people with computers that have docker limitations.
If that is incorrect then I'll have to redo this.

# Expectations
## Tests

I think I would expect the tests to be the same. I would just expect as the 
explanation suggests that a handful of test would be selected to be compiled 
natively.

## Code

I would expect the code the just have an if statement based on the flag and
either run all of the command through podman or run all of the commands that
podman would normally run on the developers computer. I don't expect there to
be much complication with that. There are only two options, native or podman.

There might be better error handling, but I am guessing that would be a fair 
deal of work and would be outside the scope of this feature.

I would expect it to look like what we had before podman.

This is from a bit of spoilers that I saw while I was trying to figure out what
this was but I would also expect now that the GitHub Actions would have to
install all of the dependencies again. I will be interested to see how that is
handled so that we aren't installing all of the dependencies all of the time. Or
maybe we are doing that. If so that would be maybe not the most efficient things
for our tests...

bdemann:
# Old Version
```javascript
import { execSync, IOType } from 'child_process';
import {
    GLOBAL_AZLE_RUST_DIR,
    GLOBAL_AZLE_RUST_BIN_DIR,
    GLOBAL_AZLE_TARGET_DIR,
    time
} from './utils';

export function compileRustCode(
    canisterName: string,
    canisterPath: string,
    stdio: IOType
) {
    execSync(
        `cd ${canisterPath} && ${GLOBAL_AZLE_RUST_BIN_DIR}/cargo build --target wasm32-wasi --manifest-path canister/Cargo.toml --release`,
        {
            stdio,
            env: {
                ...process.env,
                CARGO_TARGET_DIR: GLOBAL_AZLE_TARGET_DIR,
                CARGO_HOME: GLOBAL_AZLE_RUST_DIR,
                RUSTUP_HOME: GLOBAL_AZLE_RUST_DIR,
                CARGO_REGISTRIES_CRATES_IO_PROTOCOL: 'sparse'
                // TODO this allows changing the stack size, could be useful for stack overflow or heap out of bounds errors
                // RUSTFLAGS: '-C link-args=-zstack-size=2000000000'
            }
        }
    );

    const wasmTargetFilePath = `${GLOBAL_AZLE_TARGET_DIR}/wasm32-wasi/release/canister.wasm`;

    execSync(`cp ${wasmTargetFilePath} ${canisterPath}`);

    execSync(
        `cd ${canisterPath} && ${GLOBAL_AZLE_RUST_BIN_DIR}/wasi2ic canister.wasm ${canisterName}.wasm`,
        {
            stdio,
            env: {
                ...process.env,
                CARGO_TARGET_DIR: GLOBAL_AZLE_TARGET_DIR,
                CARGO_HOME: GLOBAL_AZLE_RUST_DIR,
                RUSTUP_HOME: GLOBAL_AZLE_RUST_DIR
            }
        }
    );
}
```

So in this old version we have three commands we are running
1. `cd <canister-path> && cargo build <with target, manifest, release, and env>`
2. `cp <wasm> <canister>`
3. `cd <canister-path> && wasi2ic canister.wasm <cansiterName>.wasm <with env>`

There is a bunch more stuff to figure out all the right paths and file names but
that's about it.

src/compiler/index.ts
```diff
@@ -68,14 +69,16 @@ async function azle() {
...
 
-            prepareDockerImage(
-                stdioType,
-                dockerImageName,
-                dockerImagePathTar,
-                dockerImagePathTarGz,
-                dockerContainerName,
-                wasmedgeQuickJsPath
-            );
+            if (nativeCompilation === false) {
```
bdemann, line 72: Who's responsibility is it to prove that this and the actual compilation are the only things that need to behave differently because of the the nativeCompilation flag? Yours as the developer or mine as the reviewer?

src/compiler/compile_rust_code.ts
```diff
@@ -1,6 +1,19 @@
 import { execSync, IOType } from 'child_process';
 
 export function compileRustCode(
+    dockerContainerName: string,
```
bdemann, line 4: It's a little confusing to have this be a required parameter when it is only used when `nativeCompilation === false`

src/compiler/get_names.ts
```diff
@@ -41,6 +41,8 @@ export async function getNamesBeforeCli() {
         .toString()
         .trim();
 
+    const nativeCompilation = process.argv.includes('--native-compilation');
```
bdemann, line 44: Could you create an issue to add `azle --help` to give a simple overview of all of the flags azle has?

src/compiler/compile_rust_code.ts
```diff
@@ -37,3 +50,15 @@ export function compileRustCode(
         { stdio }
     );
 }
+
+function compileRustCodeNatively(canisterName: string, stdio: IOType) {
+    execSync(
+        `CARGO_TARGET_DIR=target cargo build --target wasm32-wasi --manifest-path .azle/${canisterName}/canister/Cargo.toml --release`,
```
bdemann, line 56: We aren't `cd`ing or `cp`ing any more... what's up with that?

bdemann (changes requested):
# Results

There are a bunch of files that added target back to the .gitignore and added
the --native-compilation flag to the dfx.json. I don't think those files need to
be reviewed much.

test.ymal adds back all of the dependencies for native installation. It doesn't
look like it cares if the node is going to be running a native compilation test
or a podman test. Is it even possible to determine that before the test is running?

That leaves the meat of the review to pass through the rubric.

## Rubric

    1. No new files, new variable (`nativeCompilation`) looks good. New
    functions (`compileRustCodeNatively` and `compileRustCodeWithPodman`) look
    good. There isn't much more to look at here. It's just a couple of if
    statements and passing variables through.
    1. As far as readability goes I am finding it a bit confusing that we still
    use things based off of the dockerContainerName or dockerHash even if we are
    compiling natively. like [here](https://github.com/demergent-labs/azle/pull/1700/files#diff-0e6b499862c9cd05216d74349a64eec588bf77ccba21548e851162c346dc8162R25) or [here](https://github.com/demergent-labs/azle/pull/1700/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R300).
    I get that it's nice to have one function for compileRustCode and then it
    determines which way to compile it, but I would find it weird if I ever, for
    example, called it in a context where we were only compiling natively, and we
    don't even have a dockerContainerName. Realistically that will probably
    never happen...
    1. Looks good, not much has been added that would affect this, both
    `compileRustCodeWithPodman` and `compileRustCodeNatively` are encapsulated
    in their own functions, that's good.
    1. I think the only things that pertain to this category are more
    appropriately covered in the following.
    1. I guess
    https://github.com/demergent-labs/azle/pull/1700/files#diff-bc0edaf897df65b29139d6cc26dfacd8dd366af89f38b99e6e22c672df55cfe6R72
    would raise the question of if there are other things we ought to ignore if
    we are doing native compilation. Which one of us has the burden to prove
    that it's just preparing the image and then the actual compilation?
    1. This looks like a different compilation process than before?
        1. Expected
        ```bash
            cd <canister-path> && cargo build <with target, manifest, release, and env>
            cp <wasm> <canister>
            cd <canister-path> && wasi2ic canister.wasm <cansiterName>.wasm <with env>
        ```
        1. Actual
        ```bash
            cargo build <with target, manifest, release>
            wasi2ic canister.wasm <cansiterName>.wasm
        ```
        We aren't `cd`ing or `cp`ing any more... what's up with that?
    1. No updates to the tests besides what I was expecting, (a few that
    have the addition of the native-compilation flag and that's it)
    1. Where would documentation for this feature be?
        1. We should at least add this flag to `azle --help` We should add
        `azle --help`. Could you make an issue to create `azle --help` that will
        explain all of our flags and usage to developers?

## Code

Code is more or less as I was expecting, no big changes need.

## Tests

Tests are mostly unchanged with just a few changing their compilation method,
just as expected.

.github/workflows/test.yml
```diff
@@ -243,6 +243,28 @@ jobs:
                   path: /home/runner/.config/azle
                   # path: $HOME/.config/azle
                   key: config-azle-${{ hashFiles('src/compiler/Dockerfile', 'src/compiler/rust/**') }}
+              # This is for the --native-compilation tests
```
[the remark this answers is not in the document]
lastmjs: Looks like it adds ~35 seconds, technically there is a way to do this of course, but I'm not sure of a nice way to do this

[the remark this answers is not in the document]
lastmjs: I would prefer to not take the time to figure that out right now

src/compiler/compile_rust_code_with_candid_and_compiler_info.ts
```diff
@@ -11,13 +11,21 @@ export function compileRustCodeWithCandidAndCompilerInfo(
...
+    nativeCompilation: boolean
 ) {
     // This is for the Rust canister to have access to the candid file
     writeFileSync(rustStagingCandidPath, candid);
 
     // TODO why not just write the dfx.json file here as well?
     writeFileSync(compilerInfoPath, JSON.stringify(compilerInfo));
 
-    compileRustCode(dockerContainerName, canisterName, stdioType);
+    console.log('before compileRustCode nativeCompilation', nativeCompilation);
```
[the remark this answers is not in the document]
lastmjs: Removed

src/compiler/index.ts
lastmjs: Maybe I don't understand, let's discuss

src/compiler/compile_rust_code.ts
lastmjs: Sure, but we always have the dockerContainerName...wouldn't trying to add this in make things maybe unnecessarily complicated?

src/compiler/compile_rust_code.ts
lastmjs: I can tell you

bdemann (approved):

bdemann (approved):

PR #1701, opened 2024-03-10
Better logs

lastmjs:


bdemann:
# What is the feature?
Better logs

## Does it have an issue?
No

## What is it doing?
### Explanation
I'm guessing it's just better logs

### Questions
None

# Expectations
## Tests

Tests shouldn't change at all, unless it's logging during a test.

## Code

Code should have better logs, free from grammatical and spelling errors. They
should make sense and be logged at the right level.

bdemann:
I would like the record to show that I am not a fan of separate PRs that address comments from other still open PRs

bdemann (approved):
# Results
## Rubric

    1. Good. We are mapping, we have the if else. No problems here
    1. It's not the easiest code I have ever read... but I don't know what I
    would do any differently
    1. It's fine
    1. It's fine
    1. It's fine
    1. Shouldn't need any changes here.

## Code

Code is fine, and meet my expectations

## Tests

Tests are not necessary

PR #1705, opened 2024-03-11
Tfjs

lastmjs:
As an additional test for `file://` in fetch, and to show a solution to a problem that a dev has wanted since ETH Denver, I have created a TensorFlow.js example that loads a basic model. The loading of TensorFlow.js uses fetch and `file://` under-the-hood. We weren't able to easily figure out any other way of doing this, and thus the motivation for implementing the `file://` protocol in fetch.

It is still somewhat unclear how TensorFlow.js has been working in Node, considering from our research it seems `file://` is not yet implemented. Perhaps they have their own or some kind of modified library version of fetch that they've been using.

The TensorFlow.js example is not completely finished, as the model does not produce inference. I am okay with that for now, it at least loads the model and goes to the filesystem for that. In the future we can improve this example or create other TensorFlow.js examples with more advanced AI use cases.


bdemann (approved):

examples/tfjs/src/api.ts
```diff
@@ -0,0 +1,35 @@
...
+        isTypedArray: ((array: any) => {
+            return (array instanceof Float32Array ||
+                array instanceof Int32Array ||
+                array instanceof Uint8Array ||
+                array instanceof Uint8ClampedArray) as any;
+        }) as any
+    };
+
+    const model = await tf.loadLayersModel('file://spam/model.json', {
+        fetchFunc: fetch
```
bdemann, line 24: Do we need this still?

bdemann (changes requested):

examples/tfjs/src/api.ts
lastmjs: Attempting to remove

lastmjs: It was removed

Issue #1698, opened 2024-03-08
Azle 0.21.0 Documentation


lastmjs:
- [ ] Put in limitations: CORS is not respected
- [ ] add all new express examples to the examples section and hello_world readme
- [ ] update documentation about dfx, I believe the path stuff has now changed...I only had to restart my terminal, not sure what others might need to do. We can probably just...maybe say nothing because the installation has helpful output
- [ ] update dfx path stuff documentation
- [ ] explain that fd_readdir does not work
- [ ] add authentication documentation
- [ ] document that the Headers object cannot be used for authentication
- [ ] document that the following will not work: cache, keepalive and everything else that might not work
- [ ] document that the fetch is designed currently for the browser and not for nodejs
- [ ] explain that many of the default headers that the browser might attach will be missing

PR #1706, opened 2024-03-12
Backend fetch http

lastmjs:
The purpose of this PR is to provide general-purpose `http` capabilities from `fetch`. The developer should not need to understand that they are on ICP to do basic `GET` and `POST` requests. These are currently the only types of requests that are supported. I'm not sure how important it is to provide nice errors if the user tries to use unsupported methods, headers, etc...they will still get an error by the system at least for the methods I would guess.

They shouldn't have to do anything ICP-specific as I said, unfortunately there are some settings on http requests (the management canister is the underlying mechanism to actually perform these requests) that the developer must set. So we have come up with a global method on `ic` to set all of these settings. They can call this before making http requests to set those settings.

If they don't call this, they will use default settings that could be very expensive and perhaps not calibrated correctly. I believe I have the default settings pretty good for most use cases, as in it should work and not throw errors.

Generally they should set the global settings. We would like to push DFINITY/ICP to remove the need to specify these extra settings.

I have some basic tests, but maybe now or maybe in another PR it would probably be good to add a more robust set of tests.

This is a basic implementation, I'm not sure we should go for the entire fetch API right now, but the happy path. For example on the frontend fetch we already decided not to implement accepting the Request object, we might want to do the same here.


bdemann:
# What is the feature?
Adding fetch http to the backend

## Does it have an issue?
No. Looks like its a bullet point of https://github.com/demergent-labs/azle/issues/1691

## What is it doing?
### Explanation
I would expect this to be pretty similar to the front end, in that we should
need to figure out that it's just a plain http request and pass it into the
original fetch? No that's not right, on the backend we don't have an original
fetch. We have to do all of the http calls through the management canister.
So I would expect this to just be an adapter for the http protocol in fetch to
the ic.call.

So all of those weird header things that we were worrying about we will also
have to worry about here. So I guess before really diving into this I ought to
familiarize myself with fetch and the http protocol it implements.

### Questions
- [x] What is the api for the management canister http request?
- [x] How does that translate from the fetch?
    - [x] Request object?
    - [x] Options?
    - [x] How are headers represented by fetch?
        - [x] Headers object, or an object literal with string values.
- [ ] Do they have the same return type?
    - [ ] If not how do they translate?
- [x] Are there things that fetch supports that the management canister does not
    - [x] visa versa
- [x] How do we handle the cycles cost?
- [x] What does it mean that it will return an HTTP response, possibly after a
transformation. What on earth is that transformation and why do we need it?
- [x] What can we learn from the [interface spec](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-http_request)?
- [x] How are we going to handle all of the other http methods besides GET, POST, and HEAD?
- [x] What is the max_response_bytes?
- [x] What is the transform?
- [x] Why would you want to specify a max less than the 2MB max?
    - [x] if its to decrease the cycles cost why isn't that calculated
    automatically?
        - [x] Yes the price will be based off this value.
        - [ ] So why isn't that calculated automatically based off the actual
        response bytes?
- [ ] Why is the developer responsible for sending the cycles amount in the call?
    - [ ] what if they don't know the price?
    - [ ] what happens if they don't do enough?
        - [ ] does it fail?
        - [ ] if it fails do they loose all of those cycles?
        - [ ] I'm just imagining a situation where it cost 1 more cycle then what they specified and they loose it all of lack of one more cycle.
- [ ] Do we need to handle the 8192 url max? Or will we just let rust handle that
for us when we pass it the url?
    - [ ] Same question but for the limit of 64 headers
    - [ ] Same question but for 8 KiB limit on header names and values
    - [ ] Same question but for the 48 KiB limit on all total header names and values
    - [ ] Same question but for IPv4 addresses
- [x] The transform function must be exported by the canister. How do we handle that
in azle?
- [x] Cycles to pay for the call must be explicitly transferred with the call,
i.e., they are not deducted from the caller's balance implicitly (e.g., as for
inter-canister calls).
    - [ ] why not?
- [x] Is fetch still experimental in node.js? It was in 2022?
    - [x] No, in Node 21 global fetch is now stable
- [ ] What are these options
    - [ ] What is subresource integrity value??
    - [ ] What is the request mode?
    - [ ] All of the other ones I don't understand?

### Answers
#### What is the api for http requests?
Under the hood we will be calling the management canister through the rust api.
Under the management canister it has the
[http_request](https://docs.rs/ic-cdk/latest/ic_cdk/api/management_canister/http_request/index.html) that we will be using. It has some structs and enums
that will come into play later I'm guessing. Lets start with http_request (I
don't think we will be using http_request_with_closure at all).

Basically http_request makes an http request. Pretty straight forward. Let's see
The CanisterHttpRequestArgument is probably going to be what we need to look at.
We should also be aware of the cycles cost for this call. I have added a question
about how we are going to address that.

I am also told to look at the [interface spec](https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-http_request) for more information.

We will take a look at that later.

##### CanisterHttpRequestArgument

This has
- a string url, that's fine
- max_response_bytes, what is that?
- method, I'm guessing that's an Enum with POST, GET, etc. Hmm it is but it only has POST, GET, and HEAD...
- A list of headers
- an optional body of bytes
- and a transform, what is that?

I am hoping that all of that will make more sense after reading the interface spec, so let's jump to that now.

#### What can we learn from the Interface Spec
The canister should ait to issue idempotent requests. I think that's on our
developers and not on us, so I'm going to disregard that. Although that is a
troubling limitation that I hadn't thought of before...

Ah, so the transform function is used to sanitize a response so that things
like timestamps that would otherwise make two identical requests have different
responses can be removed. Makes sense, but also makes me further concerned by
the limitations of making http requests from a canister. Making them to a
canister doesn't seem too bad mind you, but if your server relies on another
server... that makes me nervous...

Oh, and the canister will only have access to the request after it's sanitized,
which I guess makes sense because if you have a replicated function working on
different date than it's other replicated parts then that would mess with the
replication.

The spec owns up to only supported GET, HEAD, and POST methods. With the
implication that that is only how it is right now and there might be support in
the future.

For security reasons only https is supported.

The max size for a request or response is 2_000_000B. That makes sense, it's
the same as the message size limit. You can configure the size of the response
to be less than 2 MB but I still don't know why you would want to do that. I
think I remember that Jordan mentioned that the price will be linked to that
size. But why not just have the price be calculated based on what is actually
returned?

Cool, I got a lot more questions from that... Let's keep plugging along on those

#### What does the fetch api look like for the http protocol?
In a simple example it looks like this
```javascript
const response = await fetch(
    "http://url.domain",
    {
        method: "POST",
        headers: {
            "My-Header-Name": "My-Header-Value"
        },
        body: [1, 2, 3, 4, 5, 6, 7]
    }
);
```
But I'm guessing I'm going to need more that just a little easy start up demo
to make sure we are adequately implement the api.

I recently learned that [undici](https://github.com/nodejs/undici) is the
project the implement the global fetch for node. Let's see if they have any
documentation that could help inform this. Secifically they have a [common API
methods](https://github.com/nodejs/undici?tab=readme-ov-file#common-api-methods)
section in their readme that will probably be helpful.

For right now I would expect us only to implement the most common API features.
And even then I don't know if it is expedient for us to do all of those right
away as our highest priority.

Even more specifically they have a section about
[fetch](https://github.com/nodejs/undici?tab=readme-ov-file#common-api-methods)
which has links to [mozilla developer
docs](https://developer.mozilla.org/en-US/docs/Web/API/fetch) and the [whatwg
specs](https://fetch.spec.whatwg.org/#fetch-method) for fetch. That's probably
an even better resource to check out. Let's move over there.

##### Syntax
```javascript
fetch(resource)
fetch(resource, options)
```

resource is either a url (a string or anything with a stringifier (included URL) that will resolve to a url essentially) or a Request object.

And then options is an object with a bunch of options.

If I can just look through what the Request object is and briefly cover all of
the options I think we'll be set. I'm guessing there is going to be a fair
amount of overlap.

##### Request Object
It has 14 properties.
- body
- if the body was used in a request yet
- cache mode (default, reload, no-cache)?
- credentials?
- destination (or content type)?
- headers
- subresource integrity value??
- method (POST, GET, etc)
- mode (cors, no-cors, same-origin, navigate)?
- mode for how redirects are handled?
- referrer?
- referrer policy?
- signal?
- url
- Those not part of options
    - bodyUsed
    - destination
    - url

I think most of those are pretty straight forward (except the ones with ?)...
So most of them aren't...

I guess I would expect our code to be able to parse all of that and add it to
the http_request appropriately, if applicable

##### Options
- That are part of the request body
    - body
    - cache
    - credentials
    - headers
    - integrity
    - method
    - mode
    - redirect
    - referrer
    - referrerPolicy
    - signal
- That are not
    - browsing topics
    - keep alive?
    - priority

At this point I am feeling a little overwhelmed with each of these. So I am
going to put them aside for now and look at each one as we get to it in the
review.



# Expectations
## Tests

- Do I want to see each option represented in at least one test?
- I might want to see the tests for each of ic.call and icp protocol and http
protocol
- I want to see a test with a url string URL object and a Request object.
- I want to see a test with a transform function
- I want to see a test with each of the three allowed methods
- Representatives of the different header approaches
- Maybe I want to see test that hit the url length max and header maxes


## Code

Code should
- have a good interface for adding a transform function
- should alert the user if they try and use something besides GET, HEAD, and POST
    - either through a warning, or error. Probably an error and throw I don't
    see how you guess what the right header to use is. I guess I would expect us
    to leverage typescript to have a type that will warn them before compilation
    that it's only those three headers that are allowed.
- only allow https calls
- export the transform function. Or require the developer to do so
- Implement a good selection the most common api features of node's global fetch
- Be able to handle a request object
- Be able to handle all of the options
    - Even if we don't support them I would expect them to be handled
- Handle both headers in a header object and object with string literals
- some way to handle cycles costs and max response size. Based on design
discussions I am expecting that they will be optional fields which if left blank
will charge the max. Which after reading the ic spec I agree is the correct move
since it is the move that they themselves make.
- I expect the code or documentation should enforce or inform the developer that
their transform function needs to be exported
- Maybe there is handling for the 8192 url max
    - Same expectation for the limit of 64 headers
    - Same expectation for 8 KiB limit on header names and values
    - Same expectation for the 48 KiB limit on all total header names and values
    - Same expectation for IPv4 addresses

bdemann:
## Responses
I almost forgot to look at responses
The response for fetch is going to be made like this
```javascript
new Response()
new Response(body)
new Response(body, options)
```

body is either null (by default) or one of
- blob
- Array Buffer
- Typed Array
- DataView
- Form Data
- ReadableStream
- URL Search Params
- String
- string literal

options are
- status (default is 200)
- status text (such as "OK" default is "")
- headers (by default is empty, is (like request) made of either a Headers object or object literal with string key/value pairs).

In rust for http_request the response looks like
a body of just bytes
a list of Headers
a Nat for status

Headers are going to be like this
```rust
pub struct HttpHeader {
    pub name: String,
    pub value: String,
}
```

I would expect the code to also properly covert from rust response to http response.

I think I would expect it to always be a blob body, with a Headers object, and
a status.

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
-    if (url.protocol === 'http:') {
-        // TODO http call to the management canister
+    if (url.protocol === 'https:' || url.protocol === 'http:') {
+        const body = await prepareRequestBody(init);
+        // TODO also do headers and method and everything like on the client?
+
+        const response = await azleFetch(`icp://aaaaa-aa/http_request`, {
+            body: serialize({
+                args: [
+                    {
```
bdemann, line 115: Should we be supported more of the options then this? Are these the same options that we talked about on the front end fetch and decided not to support?

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
             }
-        };
+        } as any;
     }
 
-    if (url.protocol === 'http:') {
-        // TODO http call to the management canister
+    if (url.protocol === 'https:' || url.protocol === 'http:') {
+        const body = await prepareRequestBody(init);
+        // TODO also do headers and method and everything like on the client?
```
bdemann, line 110: Should this be addressed before merging in?

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
+
+        const response = await azleFetch(`icp://aaaaa-aa/http_request`, {
```
bdemann, line 112: Do we call http_request vs http_request_update depending on the method? Or is it always http_request and it upgrades it's self if needed?

src/lib/fetch.ts
```diff
@@ -4,20 +4,30 @@ import { URL } from 'url';
...
-export async function azleFetch(input: any, init?: any): Promise<any> {
+export async function azleFetch(
+    input: RequestInfo | URL,
+    init?: RequestInit | undefined
+): Promise<Response> {
     if (process.env.AZLE_TEST_FETCH === 'true') {
         console.log('azleFetch has been called');
     }
 
+    if (typeof input !== 'string') {
```
bdemann, line 15: I don't think this is necessary. URL objects are pretty easy to turn into strings and it's pretty easy to sort a Request object into a url and most of the options object.

src/lib/fetch.ts
bdemann: If we aren't going to support them should be output warnings if people try to use them?

src/lib/fetch.ts
```diff
@@ -4,20 +4,30 @@ import { URL } from 'url';
...
+        throw new Error(`azleFetch input can only be of type string`);
+    }
+
     const url = new URL(input);
 
     if (url.protocol === 'icp:') {
         const canisterId = url.hostname;
         const canisterMethod = url.pathname.replace('/', '');
 
         const { method, headers, body } = init ?? {};
```
bdemann, line 25: It has come to my attention that headers is not used after this point. That seems like a problem to me

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
+                        url: input,
+                        max_response_bytes: getHttpMaxResponseBytes(),
+                        method: getHttpMethod(init),
+                        headers: init?.headers ?? [],
+                        body,
+                        transform: getHttpTransform()
+                    }
+                ],
+                cycles:
+                    globalThis._azleOutgoingHttpOptionsCycles ?? 3_000_000_000n // TODO this seems to be a conservative max size
```
bdemann, line 125: could we put this in a function getCycles() like getHttpMaxRepsonseBytes()?

bdemann: It seems like we should be able to calculate the cycles based off the getHttpMaxResponseBytes() right?

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
-        };
+        } as any;
     }
 
-    if (url.protocol === 'http:') {
-        // TODO http call to the management canister
+    if (url.protocol === 'https:' || url.protocol === 'http:') {
+        const body = await prepareRequestBody(init);
+        // TODO also do headers and method and everything like on the client?
+
```
bdemann, line 111: There are also a number of constraints on the url and headers. Do we need to enforce those? Or would us enforcing them be an opportunity to have a better error message for our users?
- 8192 url max
- No IPv4 addresses
- 64 headers
- 8 KiB limit on header names and values
- 48 KiB limit on all total header names and values

examples/ethereum_json_rpc/src/index.ts
```diff
@@ -59,46 +55,26 @@ export default Canister({
...
-                ],
-                cycles: 50_000_000n
+        ic.setOutgoingHttpOptions({
+            maxResponseBytes: 2_000n,
+            cycles: 50_000_000n,
+            transformMethodName: 'ethTransform'
+        });
+
+        const response = await fetch(url, {
+            method: 'POST',
```
bdemann, line 65: In one of the example could we have a test for a HEAD method?

bdemann: It might not be necessary it depends a lot on what we do to resolve the TODO on line 110 of fetch.ts

src/lib/fetch.ts
bdemann: Also whatever you do here keep in mind that the icp protocol currently disregards headers and methods

bdemann, line 110: I would like to see some tests that send headers and have those headers verified so we can catch things like not getting them hooked up

src/lib/fetch.ts
bdemann, line 125: Do we have any calculations that help us arrive at 3_000_000_000n or is that just a best guess? or found through trial and error?

src/lib/fetch.ts
bdemann, line 15: If we add support for URL and Request object I would love to see some tests for it.

src/lib/fetch.ts
bdemann, line 111: If we add constraints to the url and headers we should test the edge cases around those limits

bdemann (changes requested):
# Results
## Rubric

    - There are a handful of new `any` types.
    - return value for the http protocol is a little hard to read at first glance
        - could be more declarative
    - azleFetch is starting to get pretty long, could we break it up into smaller functions
        - slitting along protocol lines seems like a good natural breaking point
    - I like the updates to getBalance and getBlockByNumber in the ethereum json rpc example
    - I like the reuse of azleFetch in the http protocol implementation
    - Discussed in detail below
    - Discussed in detail below
    - Discussed in detail below

## Code

Lets start with responses so we don't forget them again. Responses are not
quite what I was expecting, I have added a comment to that effect. Overall
I do like the http implementation, just using the icp protocol to call the
management canister as we had been doing before. I guess on that note we used
to have responses from azleFetch that were never like a Response object and it
seemed to work then ¯\\\_(ツ)\_/¯

I like the api for adding a transform function. I'm just a little nervous about
how to best teach people what a transform method is, why they need it, and
making sure that they have added it as a query or update method.

There was a TODO to address unsupported methods. I think that should be pretty
easy to add and would be beneficial. Or at least as beneficial as the odds of
someone doing a method besides POST or GET.

It allows http calls. It shouldn't

### Exporting transform functions
Knowing what I now do about transform functions the developer must provide one
if one is needed. That is not something we can do. Hopefully they don't need
one, but what happens if they do and they don't provide one, or they make one,
but they don't add it as a query method?

### Implement a good selection the most common api features of node's global fetch
I wrote this one before diving into the mozilla developer docs. I'm not sure it
actually applies like I thought it would when I wrote it. because the protocol
is pretty straight forward. We would just need to make sure we cover all of the
options. Or that we can pass the buck to the management canister for not
supported all of the options.

What options do we support?
It looks like almost none of them. The common ones were
- [x] body
- [ ] cache
- [ ] credentials
- [x] headers
- [ ] integrity
- [x] method
- [ ] mode
- [ ] redirect
- [ ] referrer
- [ ] referrerPolicy
- [ ] signal

We don't currently support URL or Request objects for the fetch input, but I
think those would be very easy to add support for, especially given that we just
ignore the options that aren't body, headers, and method.

## Tests

- Do I want to see each option represented in at least one test?
    - Right now options aren't supported so I won't expect these until we add
    them
- I might want to see the tests for each of ic.call and icp protocol and http
protocol
    - Actually I don't think we need this because http is calling the icp
    protocol so leaving it as is should provide this coverage.
- I want to see a test with a url string URL object and a Request object.
    - I think we should support this and add tests for it.
- I want to see a test with a transform function
    - We already have that in both of the examples that this pr touches.
- I want to see a test with each of the three allowed methods
    - We have GET and POST tests, but no HEAD tests.
- Representatives of the different header approaches
    - We don't have any tests for headers and consequently we didn't catch
    that they are being completely disregarded
- Maybe I want to see test that hit the url length max and header maxes
    - We don't currently enforce those limits, if we don't then it's probably
    not important, and would be on dfinity to test those, if we do then we
    should test them.


## Misc Concerns

It seems like doing a canister side fetch call is going to very quickly start to
require some deep IC knowledge.
    - Transform functions (which brings query/update functions (which brings candid)), cycle costs, replication and idempotent requests

src/lib/fetch.ts
bdemann: http_client/fetch_ic/call.ts has a great function getUrlString that could be moved to a place accessible by both files and then used here. Something like that would be great! Just don't forget to also look for the options in that Request object

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
...
+                        transform: getHttpTransform()
+                    }
+                ],
+                cycles:
+                    globalThis._azleOutgoingHttpOptionsCycles ?? 3_000_000_000n // TODO this seems to be a conservative max size
+            })
+        });
+        const responseJson = await response.json();
+
+        return {
```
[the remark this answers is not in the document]
bdemann: We have to make it because it doesn't come from QuickJS, we might do this in the future but not now

src/lib/fetch.ts
```diff
@@ -125,3 +167,107 @@ function determineCandidPath(canisterId: string, candidPath?: string): string {
...
+            get: null
+        };
+    }
+
+    return {
+        [init.method?.toLowerCase()]: null
+    };
+}
+
+function getHttpTransform() {
```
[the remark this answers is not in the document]
bdemann: dfx should give them an error, we could polish it more, but we can do that later.

src/lib/fetch.ts
bdemann: the management canister just has http_request, other canisters are the ones with http_request and http_request_update

src/lib/fetch.ts
bdemann: please add URL and add a comment about not adding Request

src/lib/fetch.ts
bdemann: Can you underscore these to signify that they are unused on purpose

src/lib/fetch.ts
bdemann: We've decided it would be best just let the IC enforce this, but it would be good to have this in the documentation. Please make an issue to address this when we do documentation

examples/ethereum_json_rpc/src/index.ts
bdemann: yes please do this, thank you!

src/lib/fetch.ts
bdemann: We are not going to do that, don't worry about this

src/lib/fetch.ts
lastmjs: I just removed them

src/lib/fetch.ts
```diff
@@ -91,11 +102,42 @@ export async function azleFetch(input: any, init?: any): Promise<any> {
             text: async () => {
                 return contents.toString();
             }
-        };
+        } as any;
     }
 
-    if (url.protocol === 'http:') {
-        // TODO http call to the management canister
+    if (url.protocol === 'https:' || url.protocol === 'http:') {
```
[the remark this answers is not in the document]
lastmjs: Okay I got rid of it actually FYI

src/lib/fetch.ts
```diff
@@ -125,3 +167,107 @@ function determineCandidPath(canisterId: string, candidPath?: string): string {
...
     );
 }
+
+function getHttpMaxResponseBytes() {
+    return globalThis._azleOutgoingHttpOptionsMaxResponseBytes === undefined
+        ? []
+        : [globalThis._azleOutgoingHttpOptionsMaxResponseBytes];
+}
+
+// TODO throw errors on unsupported methods?
```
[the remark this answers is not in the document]
lastmjs: I decided to throw on unsupported methods, FYI

src/lib/fetch.ts
lastmjs: I went through and added the same warning on the backend, there are a couple properties that might exist that we don't have warnings for, but the types are off...I'm willing to let that slide for now. I also added a warning for the window property to the frontend and backend fetch

src/lib/fetch.ts
lastmjs: I have added support for headers and I have made it so that only GET, HEAD, and POST are supported

src/lib/fetch.ts
lastmjs: This restriction is gone, I made a new function called `getUrl` and added the comment about Request there

src/lib/fetch.ts
lastmjs: I have created the function but I have not yet tried to calculate cycles automatically, I will attempt to do that

src/lib/fetch.ts
lastmjs: I have added a checkbox for this here: https://github.com/demergent-labs/azle/issues/1698

src/lib/fetch.ts
lastmjs: We aren't going to add support for Request, but I will add tests for URL

src/lib/fetch.ts
lastmjs: I have a cycles calculation, it works and is accurate according to the documentation, but ~20x more expensive than it should be according to the errors from the IC itself ($.02 per call instead of 20x less). It's just a default, I have asked in the forum to see if we can figure out why this is the case: https://forum.dfinity.org/t/a-new-price-function-for-https-outcalls/20838/21

lastmjs: I have a cycles calculation, it works and is accurate according to the documentation, but ~20x more expensive than it should be according to the errors from the IC itself ($.02 per call instead of 20x less). It's just a default, I have asked in the forum to see if we can figure out why this is the case: https://forum.dfinity.org/t/a-new-price-function-for-https-outcalls/20838/21

lastmjs: Apparently I have it correct, the price I am calculating is the 13 node subnet price...we'll just leave it hardcoded like this for now then, I have added the ability to set the subnet size with the global http call settings (or I will add that)

src/lib/fetch.ts
lastmjs: Done

src/lib/fetch.ts
lastmjs: I've added one test for URL

src/lib/fetch.ts
lastmjs: I added the ability to change subnet size

bdemann (approved):

PR #1711, opened 2024-03-19
get rid of the need for explicitly exporting Server

lastmjs:
The developer is no longer required to explicitly export Server, they can create Express or other Node.js server apps by just calling listen on their server.

There is one failing test, new. This is expected as new tests on a previous version of Azle, and that previous version is incompatible with this new version...something like that, it's most likely fine and expected.

### DIFF
--- a/examples/fs/src/index.ts
+++ b/examples/fs/src/index.ts
@@ -1,4 +1,3 @@
-import { Server } from 'azle';
 import express, { Request } from 'express';
 import {
     createReadStream,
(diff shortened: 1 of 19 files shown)


examples/fs/src/index.ts
bdemann, line 1: I'm assuming this file is just getting rid of the Server export and the diff just got weird. If that's the case just resolve this comment and move on. If not let me know and I'll come back and have a closer look

src/compiler/compile_typescript_code.ts
bdemann, line 1: On problem I have is that an important part of this update happened [here](https://github.com/demergent-labs/wasmedge-quickjs/commit/c21ff69f442998e4cda4619166e23a9bc91418be#diff-5c2b50f54aff71876721c24a02e01579b7a16a8fd5f0aa087879abbdf1f28f91R788). And it was not obviously part of this review. Could we add a link to any changes we make to forks of other dependencies that make these features work? Just for easier reviewing in the future.

src/compiler/compile_typescript_code.ts
```diff
@@ -11,34 +11,40 @@ export function compileTypeScriptToJavaScript(
...
-            };            
-
-            // TODO I do not know how to get the module exports yet with wasmedge_quickjs
-            globalThis.exports.canisterMethods = canisterMethods;
+            // TODO This setTimeout is here to allow asynchronous operations during canister initialization
+            // for Server canisters that have chosen not to use export default Server
+            // This seems to work no matter how many async tasks are awaited, but I am still unsure about how it will
+            // behave in all async situations
+            setTimeout(() => {
+                const canisterMethods = CanisterMethods.default !== undefined ? CanisterMethods.default() : Server(() => globalThis._azleNodeServer)();
```
bdemann, line 26: This was a little challenging for me to read at first. If we could make it a little more declarative that would be good

bdemann, line 26: Bonus points if you copy this stuff out of this file into a place where our auto-formatter can format it and paste it back here.

bdemann (changes requested):

src/compiler/compile_typescript_code.ts
```diff
@@ -11,34 +11,40 @@ export function compileTypeScriptToJavaScript(
             // Trying to make sure that all globalThis dependencies are defined
```
bdemann, line 11:
```suggestion
// Trying to make sure that all globalThis dependencies are defined
// Before the developer imports azle on their own
import 'azle';
import { ic, Server } from 'azle';
import { toDidString } from 'azle/src/lib/candid/did_file/to_did_string';
import {
    DidVisitor,
    getDefaultVisitorData
} from 'azle/src/lib/candid/did_file/visitor';
export { Principal } from '@dfinity/principal';
export * from './${main}';
import * as CanisterMethods from './${main}';

// TODO This setTimeout is here to allow asynchronous operations during canister initialization
// for Server canisters that have chosen not to use export default Server
// This seems to work no matter how many async tasks are awaited, but I am still unsure about how it will
// behave in all async situations
setTimeout(() => {
    const canisterMethods =
        CanisterMethods.default !== undefined
            ? CanisterMethods.default()
            : Server(() => globalThis._azleNodeServer)();

    globalThis.candidInfoFunction = () => {
        const candidInfo = canisterMethods.getIdl([]).accept(new DidVisitor(), {
            ...getDefaultVisitorData(),
            isFirstService: true,
            systemFuncs: canisterMethods.getSystemFunctionIdls()
        });

        return JSON.stringify({
            candid: toDidString(candidInfo),
            canisterMethods: {
                // TODO The spread is because canisterMethods is a function with properties
                // TODO we should probably just grab the props out that we need
                ...canisterMethods
            }
        });
    };

    // TODO I do not know how to get the module exports yet with wasmedge_quickjs
    globalThis.exports.canisterMethods = canisterMethods;
});
```

bdemann: Here is the bonus point formatted version. I mostly just put it here because I wanted to try out the suggestion feature. I'm not sure it will work as I am hoping

src/compiler/compile_typescript_code.ts
lastmjs: Sounds like a good idea

examples/fs/src/index.ts
lastmjs: I assume so, we also have tests

src/compiler/compile_typescript_code.ts
lastmjs: I would like to address this later I have made an issue for it: https://github.com/demergent-labs/azle/issues/1715

lastmjs: I would like to address this later I have made an issue for it: https://github.com/demergent-labs/azle/issues/1715

src/compiler/compile_typescript_code.ts
lastmjs: It would be nice to actually get this process to be automatic, I would like to address this later I have made an issue for it: https://github.com/demergent-labs/azle/issues/1715

bdemann (approved):

PR #1720, opened 2024-03-27
Infinite files test testing

bdemann:


bdemann:
This has served is purposes and was never meant to be merged in

PR #1728, opened 2024-03-29
add new identity to manage file uploading

bdemann:


src/compiler/file_uploader/uploader_actor.ts
```diff
@@ -1,11 +1,12 @@
 import { Actor, ActorMethod, ActorSubclass } from '@dfinity/agent';
 import { createAuthenticatedAgent } from '../../../dfx';
+import { AZLE_UPLOADER_IDENTITY_NAME } from './uploader_identity';
 
 export type UploaderActor = ActorSubclass<_SERVICE>;
 
 export async function createActor(
     canisterId: string,
-    identityName?: string
+    identityName: string = AZLE_UPLOADER_IDENTITY_NAME
```
lastmjs, line 9: Why do we default to this? Can we just make this required? Doesn't seem like we're using the default anywhere anyways?

src/compiler/file_uploader/uploader_identity.ts
```diff
@@ -0,0 +1,20 @@
+import {
+    addController,
+    generateIdentity,
+    getPrincipal,
+    identityExists
+} from '../../../dfx';
+
+export const AZLE_UPLOADER_IDENTITY_NAME = '_azle_file_uploader_identity';
```
lastmjs, line 8: Should we make this an environment variable now just in case? And default to that name?

dfx/index.ts
```diff
@@ -60,8 +60,38 @@ export function getIdentityName(): string {
     return execSync(`dfx identity whoami`).toString().trim();
 }
 
-export function generateIdentity(name: string) {
-    execSync(`dfx identity new ${name} --storage-mode plaintext`);
+type StorageMode = 'keyring' | 'password-protected' | 'plaintext';
+
+export function generateIdentity(
+    name: string,
+    storageModeCool?: StorageMode
```
lastmjs, line 67: This probably shouldn't say cool

.github/workflows/test.yml
```diff
@@ -214,6 +214,7 @@ jobs:
         env:
             ETHEREUM_URL: ${{ secrets.ETHEREUM_URL }}
             AZLE_USE_DOCKERFILE: ${{ matrix.azle_source == 'repo' }}
+            AZLE_PLAINTEXT_IDENTITY: true
```
lastmjs, line 217: Should we not let them pass in the storage mode that they want?

dfx/index.ts
```diff
@@ -60,8 +60,38 @@ export function getIdentityName(): string {
...
+        console.info();
+        return execSync(`dfx identity new ${name}`, {
+            stdio: ['inherit', 'pipe', 'inherit']
+        });
+    }
+    if (storageMode === 'password-protected') {
+        console.info(`You will be prompted to create a password for ${name}`);
+        console.info();
+    }
+    return execSync(`dfx identity new ${name} --storage-mode ${storageMode}`, {
```
lastmjs, line 83: Seems like the user won't get the warning here about being prompted for their password

dfx/index.ts
```diff
@@ -60,8 +60,38 @@ export function getIdentityName(): string {
...
+        stdio: ['inherit', 'pipe', 'inherit'] // TODO add todo here and aboe
```
lastmjs, line 84: What are these todos about?

dfx/index.ts
```diff
@@ -60,8 +60,38 @@ export function getIdentityName(): string {
...
+    });
+}
+
+function determineStorageMode(
```
lastmjs, line 88: Should we let the storage mode env var actually ser the storage mode?

And if it isn't one of the hard coded modes we can throw an error

dfx/index.ts
```diff
@@ -60,8 +60,38 @@ export function getIdentityName(): string {
...
+    storageMode?: StorageMode
+): StorageMode | undefined {
+    if (process.env.AZLE_PLAINTEXT_IDENTITY === 'true') {
+        return 'plaintext';
+    }
+    return storageMode;
 }
```
lastmjs, line 95: When do we ever use the use or remove identity functions? This seems scary and hopefully never used, I wonder if we should remove these

dfx/index.ts
```diff
@@ -92,3 +127,48 @@ export async function getIdentity(
...
+    const cmd = `dfx identity export ${identityName}`;
+    const result = execSync(cmd, {
+        stdio: ['inherit', 'pipe', 'inherit'] // TODO I would prefer it to pipe the stderr but it will fail immediately if you do that
+    })
+        .toString()
+        .trim();
+    return result;
+}
+
+export function getIdentity(identityName?: string): Secp256k1KeyIdentity {
```
lastmjs, line 145: We should probably rename this to getSecp256k1Identity, or allow passing in the type of identity, since this won't work generally

dfx/index.ts
```diff
@@ -75,11 +105,16 @@ export function getIdentities(): string[] {
...
+    const identities = getIdentities();
+    return identities.includes(name);
+}
+
 export function removeIdentity(name: string) {
     execSync(`dfx identity remove ${name}`);
 }
 
-export async function getIdentity(
+export async function getIdentityFromPemFile(
```
lastmjs, line 117: This should probably rename with secp256k1 in mind, or it should accept as a parameter the type of key, since this won't work generally

dfx/index.ts
```diff
@@ -92,3 +127,48 @@ export async function getIdentity(
     );
     return Secp256k1KeyIdentity.fromPem(await readFile(identityPath, 'utf-8'));
 }
+
+export function getPemKey(identityName: string = getIdentityName()): string {
```
lastmjs, line 131: These defaults are scary, I think we should not use them and just explicitly pass in our parameters considering the absolutely intense nature of getting this identity stuff correct

lastmjs, line 131: What does getIdentityName do? Is it grabbing the name of the current dfx identity? If so I think we should change its name to indicate that

lastmjs (changes requested):

dfx/index.ts
bdemann: done

dfx/index.ts
bdemann: If we get to this point they won't be prompted for their password since it will be either keychain or plaintext, neither of those use a password

dfx/index.ts
bdemann: done

dfx/index.ts
bdemann: I had asked about this last PR, where this was introduced and we had decided to just leave them in. If we decide that we want to overturn that decisions I will remove it, but we should probably talk about why we are changing our mind about it

dfx/index.ts
bdemann: This one isn't used anymore and is essentially the same but less powerful as the other getIdentity function. I just got rid of it instead.

dfx/index.ts
bdemann: sure

bdemann: okay

src/compiler/file_uploader/uploader_actor.ts
bdemann: sure, we do use it in the tests... but I guess it's fine

.github/workflows/test.yml
bdemann: sure

src/compiler/file_uploader/uploader_identity.ts
bdemann: sure

dfx/index.ts
bdemann: sure

dfx/index.ts
```diff
@@ -56,42 +53,45 @@ export async function createAuthenticatedAgent(
...
+function determineStorageMode(): StorageMode | undefined {
+    const mode = process.env.AZLE_IDENTITY_STORAGE_MODE;
+    if (
+        mode !== 'plaintext' &&
+        mode !== 'keyring' &&
+        mode !== 'password-protected' &&
+        mode !== undefined
+    ) {
+        throw new Error(
+            `AZLE_IDENTITY_STORAGE_MODE must be 'plaintext', 'keyring', 'password-protected', or undefined`
```
lastmjs: Remove undefined comment here

src/compiler/file_uploader/uploader_identity.ts
```diff
@@ -5,7 +5,8 @@ import {
     identityExists
 } from '../../../dfx';
 
-export const AZLE_UPLOADER_IDENTITY_NAME = '_azle_file_uploader_identity';
+export const AZLE_UPLOADER_IDENTITY_NAME =
```
lastmjs, line 8: nullish coalescing

dfx/index.ts
```diff
@@ -114,46 +114,34 @@ export function removeIdentity(name: string) {
...
 
-export function getIdentity(identityName?: string): Secp256k1KeyIdentity {
+export function getSecp256k1KeyIdentity(
+    identityName: string
+): Secp256k1KeyIdentity {
     return Secp256k1KeyIdentity.fromPem(getPemKey(identityName));
 }
 
-export function getPrincipal(identityName: string = getIdentityName()): string {
+export function getPrincipal(identityName: string = whoami()): string {
```
lastmjs: Get rid of optional

lastmjs (changes requested):

.github/workflows/test.yml
lastmjs: I'm a little concerned about the quotes now

dfx/index.ts
lastmjs: get rid of use identity

lastmjs (approved):

PR #1735, opened 2024-04-04
New eslint fixes

bdemann:
This adds fixes for 10 of the offending eslint rules reducing us down to only 7 tech-debt disabled rules.

Add 9 additional rules that were easy to fix and seem like they would be good.
- func style (function should be declarations not declarations)
- array-callback-return (map, filter, reduce, etc, should have returns (thus avoiding undefined[])
- no-template-curly-in-string (to avoid accidentally doing "This is ${myValue}" instead of `This is ${myValue}`
- eqeqeq (enforcing === instead of == and !== instead of !=)
- prefer-template (use template literal instead of string concat)
- no-param-reassign
- @typescript-eslint/prefer-for-of
- prefer-arrow-callback
- no-var

> [!CAUTION]
> Change the base of this PR to main before merging and merge https://github.com/demergent-labs/azle/pull/1734 first.


src/lib/canister_methods/execute_method.ts
```diff
@@ -13,7 +13,7 @@ export function executeMethod(
     returnCandidType: CandidType,
     manual: boolean
 ) {
```
lastmjs, line 15: Can we make T = any in encode and decode so that we don't have to explicitly pass in <any>?

src/lib/globals.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-var */
```
lastmjs, line 1: Let's put this exception on the specific lines it applies to

http_client/fetch_ic/actor.ts
```diff
@@ -91,7 +91,14 @@ export interface _SERVICE {
             upgrade: [] | [boolean];
             streaming_strategy:
                 | []
-                | [{ Callback: { token: {}; callback: [Principal, string] } }];
+                | [
+                      {
+                          Callback: {
+                              token: unknown; // TODO I am not sure if unknown is the right option here. Maybe `object`?
```
lastmjs, line 97: We need to fix the token

http_client/fetch_ic/actor.ts
```diff
@@ -110,7 +117,14 @@ export interface _SERVICE {
             upgrade: [] | [boolean];
             streaming_strategy:
                 | []
-                | [{ Callback: { token: {}; callback: [Principal, string] } }];
+                | [
+                      {
+                          Callback: {
+                              token: unknown;
```
lastmjs, line 123: Fix the token

lastmjs (changes requested):

src/lib/canister_methods/execute_method.ts
bdemann: done

http_client/fetch_ic/actor.ts
bdemann: done

http_client/fetch_ic/actor.ts
bdemann: done

PR #1740, opened 2024-04-11
Type checked eslint

bdemann:


bdemann:
Close this because we are not ready to work on this at all. We will need a lot of work before type checked linting is good to go

Issue #1749, opened 2024-04-18
get reloadjs to be authenticated now...maybe it should work in production?


lastmjs:
- [x] Let's also log the reload somehow

PR #1750, opened 2024-04-19
add authentication to autoreload, get rid of warning, add replica log…

lastmjs:
Closes https://github.com/demergent-labs/azle/issues/1749

When we first implemented autoreloading, we did not add any authentication to the canister method. So we made sure to not compile it into the canister unless the `AZLE_AUTORELOAD` environment variable was set to true, and then if deploying to non-local network we logged a warning to the user.

Now we can easily create authenticated agents using the developer's own dfx identity, and so I've added an authentication guard to the reload_js canister method. I've also changed the name of the guard to something I like better from doing the same thing in Kybra. I've also changed the wording of the error message slightly.

I've removed the warning in Azle when deploying to a non-local network. I've added a log to the replica when the JS is fully reloaded


bdemann (changes requested):
Looks good! There is one failing tests that will be critical to fix with large files

PR #1751, opened 2024-04-19
Improve terminal error output

lastmjs:
Closes: https://github.com/demergent-labs/azle/issues/1752

I've introduced an `execSyncPretty` function to be used in place of `execSync`. This will catch the default error output and create a much nicer terminal output for the developer. It has a default stack trace and should be quite practical. Not the most beautiful still, we might want to improve it in the future, but it should be much better than what currently happens.

What currently happens is confusing and verbose, it logs stdout, stderr, with a lot of Uint8Arrays or buffers or such, and it is generally scary-looking and not helpful.

That has hopefully been fixed or greatly improved with this PR.

### DIFF
--- a/src/compiler/compile_rust_code.ts
+++ b/src/compiler/compile_rust_code.ts
@@ -1,4 +1,6 @@
-import { execSync, IOType } from 'child_process';
+import { IOType } from 'child_process';
+
+import { execSyncPretty } from './utils/exec_sync_pretty';
 
 export function compileRustCode(
     dockerContainerName: string,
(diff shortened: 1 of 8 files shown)


src/compiler/compile_rust_code.ts
```diff
@@ -18,47 +20,47 @@ function compileRustCodeWithPodman(
     canisterName: string,
     stdio: IOType
 ) {
-    execSync(
+    execSyncPretty(
         `podman exec ${dockerContainerName} rm -rf /.azle/${canisterName}`,
-        { stdio }
+        stdio
```
bdemann, line 25: Here to remind my self to understand why this is just stdio and not an object

bdemann: Okay so if I wanted to do "['inherit', 'pipe', 'inherit']" I would just pass that in instead of doing {stdio: ['inherit', 'pipe', 'inherit']}. Is that right?

bdemann: No it wouldn't do that, I'll leave a comment on the actual implementation

src/compiler/utils/exec_sync_pretty.ts
```diff
@@ -0,0 +1,9 @@
+import { execSync, IOType } from 'child_process';
+
+export function execSyncPretty(command: string, stdio?: IOType): Buffer {
```
bdemann, line 3: Could we update the type of stdio to be IOType | IOType[]?

bdemann: Use case would be in dfx/index.ts

bdemann (changes requested):

src/compiler/utils/exec_sync_pretty.ts
lastmjs: Done

PR #1770, opened 2024-05-02
Sqlite

lastmjs:


examples/sqlite_drizzle/test/test.ts
```diff
@@ -0,0 +1,7 @@
+import { getCanisterId } from 'azle/dfx';
+import { getTests } from 'azle/examples/sqlite/test/tests';
```
lastmjs, line 2: So it turns out this actually is a problem in CI/CD with tests, who would have known right?

examples/sqlite_drizzle/src/entities/posts/db.ts
```diff
@@ -0,0 +1,134 @@
...
+        : {
+              id: results[0].posts.id,
+              title: results[0].posts.title,
+              body: results[0].posts.body,
+              user: results[0].users
+          };
+}
+
+export async function countPosts(drizzleDb: DrizzleDb): Promise<number> {
+    const results = await drizzleDb
```
bdemann, line 62: drizzle has a way to use the SQL COUNT function
```javascript    
    const results = await drizzleDb
        .select({
            count: count()
        })
        .from(Posts);

    return results[0]?.count ?? 0;
```

This would be more robust because if you delete posts won't it add post starting at the biggest index and not fill in empty indexes?

examples/sqlite_drizzle/src/entities/users/db.ts
```diff
@@ -0,0 +1,93 @@
...
+    drizzleDb: DrizzleDb,
+    id: number
+): Promise<User | null> {
+    const users = await drizzleDb.select().from(Users).where(eq(Users.id, id));
+
+    return users.length === 0 ? null : users[0];
+}
+
+export async function countUsers(drizzleDb: DrizzleDb): Promise<number> {
+    const results = await drizzleDb
```
bdemann, line 33: would it be better to use the count() method built into drizzle? (see similar comment in Post)

examples/sqlite/src/entities/users/db.ts
```diff
@@ -0,0 +1,92 @@
...
+        db,
+        convertUser
+    );
+
+    return users.length === 0 ? null : users[0];
+}
+
+export function countUsers(db: Database): number {
+    const results =
+        sqlite<number>`SELECT id FROM users ORDER BY id DESC LIMIT 1`(
```
bdemann, line 31: Can't you just do this?
```sql
SELECT COUNT(*)
FROM users
```

examples/sqlite/src/entities/posts/db.ts
```diff
@@ -0,0 +1,100 @@
...
+        sqlite<Post>`SELECT * FROM posts JOIN users ON posts.user_id = users.id WHERE posts.id = ${id}`(
+            db,
+            convertPost
+        );
+
+    return posts.length === 0 ? null : posts[0];
+}
+
+export function countPosts(db: Database): number {
+    const results =
```
bdemann, line 33: Can't you just do this? (Same question as in countUsers)
```sql
SELECT COUNT(*)
FROM posts
```

examples/sqlite/test/tests.ts
```diff
@@ -2,28 +2,525 @@ import * as dns from 'node:dns';
...
+function usersTestsAfterBatch(origin: string): Test[] {
+    return [
+        {
+            name: '/users not empty',
+            test: async () => {
+                const response = await fetch(`${origin}/users`);
+                const responseJson = await response.json();
+
+                return {
+                    Ok: responseJson.length === 800
```
bdemann, line 453: I'm confused, shouldn't this be 500?

bdemann: Also should we have a constant instead of magic numbering this everywhere?

examples/sqlite/test/tests.ts
```diff
@@ -2,28 +2,525 @@ import * as dns from 'node:dns';
...
+            }
+        },
+        {
+            name: '/posts/count 300',
+            test: async () => {
+                const response = await fetch(`${origin}/posts/count`);
+                const responseJson = await response.json();
+
+                return {
+                    Ok: responseJson === 300
```
bdemann, line 505: constant instead of magic number?

examples/sqlite/test/tests.ts
```diff
@@ -2,28 +2,525 @@ import * as dns from 'node:dns';
...
+                    Ok: responseJson === false
+                };
+            }
+        },
+        ...usersTestsBeforeBatch(origin),
+        ...postsTestsBeforeBatch(origin),
+        {
+            name: '/users/batch/499',
+            test: async () => {
+                const response = await fetch(`${origin}/users/batch/499`, {
```
bdemann, line 38: Could we do a constant instead of a magic number, and could we do something like FINAL_TOTAL - 1 or something to better illustrate that we added one in the before batch and we are filling it all the way up to a set number? Alternatively have the final total be BATCH_ADD_USER_COUNT + 1. (thought with 500 being the final it does feel more like a FINAL_TOTAL - 1 situation, while 501 would feel more like a BATCH_ADD_USER_COUNT + 1 situation)

bdemann (changes requested):
Other ideas for things to test
- Transactions
- Rolling back?
- Indexes
- Subqueries
- Aggregate functions like SUM, AVG, MIN, MAX
- Group by
- Applying multiple migrations

examples/sqlite/src/entities/posts/db.ts
lastmjs: Does a full table scan and runs out of cycles if I remember correctly. It wasn't straight-forward to get a count in raw sql

examples/sqlite/test/tests.ts
lastmjs: Why should it be 500?

examples/sqlite_drizzle/test/test.ts
lastmjs: I have just addressed this and it works locally.

examples/sqlite_drizzle/src/entities/posts/db.ts
lastmjs: Resolved

examples/sqlite/src/entities/users/db.ts
lastmjs: Done

examples/sqlite/src/entities/posts/db.ts
lastmjs: Done

examples/sqlite/test/tests.ts
lastmjs: It should be 800 because every post must create a user. I've added constants to make this more clear

examples/sqlite/test/tests.ts
lastmjs: Done

examples/sqlite/test/tests.ts
lastmjs: Did something like this

bdemann (approved):

PR #1774, opened 2024-05-06
Sqlite typeorm

lastmjs:


examples/sqlite/test/tests.ts
```diff
@@ -487,11 +487,11 @@ function postsTestsAfterBatch(origin: string): Test[] {
         {
             name: '/posts not empty',
             test: async () => {
-                const response = await fetch(`${origin}/posts?limit=300`);
+                const response = await fetch(`${origin}/posts?limit=200`);
```
bdemann, line 490: Why are these 200 now instead of 300? Can they be a const instead of a magic number?

bdemann (changes requested):

examples/sqlite/test/tests.ts
lastmjs: This was made a constant in the sqlite PR

PR #1785, opened 2024-05-21
Improve testing 1781

bdemann:


examples/basic_bitcoin/test/tests.ts
```diff
@@ -21,17 +21,19 @@ const FIRST_AMOUNT_SENT = SINGLE_BLOCK_REWARD / 2n;
 const SECOND_AMOUNT_SENT = SINGLE_BLOCK_REWARD * 2n;
 const TO_ADDRESS = 'n4HY51WrdxATGEPqYvoNkEsTteRfuRMxpD';
 
-let lastTxid = '';
-let toAddressPreviousBalance = 0n;
-let canisterPreviousBalance = 0n;
+type BasicBitcoinContext = {
```
lastmjs, line 24: All of these should be optional...but actually null?

lastmjs: Let's try to do the crazy context type changing thing, otherwise make them optional/undefined

examples/basic_bitcoin/test/tests.ts
```diff
@@ -90,12 +95,18 @@ export function getTests(
...
                 );
 
-                return {
-                    Ok:
-                        utxosResult.tip_height === FIRST_MINING_SESSION &&
-                        utxosResult.utxos.length === FIRST_MINING_SESSION &&
-                        checkUtxos(utxosResult.utxos)
-                };
+                return equals(utxosResult, FIRST_MINING_SESSION, {
+                    equals: (actual: GetUtxosResult, expected: number) => {
```
lastmjs, line 99: equals equals is weird

lastmjs: customEquals

examples/apollo_server/test/tests.ts
```diff
@@ -26,12 +34,7 @@ export function getTests(canisterId: string): Test[] {
                         })
                     });
                     const responseJson = await response.json();
-
-                    return {
-                        Ok:
-                            responseJson.data.books.length === 2 &&
-                            responseJson.data.books[0].title === 'The Awakening'
-                    };
+                    return equals(expectedResult, responseJson);
```
lastmjs, line 37: Actual then expected...

examples/candid_encoding/test/tests.ts
```diff
@@ -318,11 +219,13 @@ export function get_tests(
                     );
 
                 return {
-                    Ok:
-                        candid_encoded.Ok === true &&
-                        candid_decoded_result.includes('record') &&
-                        candid_decoded_result.includes('John') &&
-                        candid_decoded_result.includes('Doe')
+                    Ok: {
+                        passes:
```
lastmjs, line 223: Rename to success

examples/candid_keywords/test/tests.ts
```diff
@@ -1,5 +1,5 @@
 import { ActorSubclass } from '@dfinity/agent';
-import { Test } from 'azle/test';
+import { pass, Test } from 'azle/test';
```
lastmjs, line 2: Rename to succeed

examples/complex_types/test/tests.ts
```diff
@@ -1,8 +1,16 @@
 import { ActorSubclass } from '@dfinity/agent';
-import { Test } from 'azle/test';
+import { createTestResult, equals, Test } from 'azle/test';
 
 import { _SERVICE } from './dfx_generated/complex_types/complex_types.did.d';
 
+const USER_1 = {
```
lastmjs, line 6: Rename to EXPECTED_USER_1

examples/ethers_base/test/tests.ts
```diff
@@ -156,9 +154,14 @@ export function getTests(canisterId: string): Test[] {
                 });
                 const responseJson = await response.json();
 
-                return {
-                    Ok: BigInt(responseJson.__bigint__) < 100_000_000_000_000n
-                };
+                const expetedBalance = 100_000_000_000_000n;
```
lastmjs, line 157: Spelled wrong

test/index.ts
```diff
@@ -1,26 +1,37 @@
...
+    prep?: (context: Context) => Promise<Context | void>;
+    test?: (context: Context) => Promise<AzleResult<boolean, string, Context>>;
 };
 
 // export type Variant<T> = Partial<T>;
 
-export type AzleResult<T, E> = Partial<{
-    Ok: T;
+export type AzleResult<T, E, Context = any> = Partial<{
+    Ok: { passes: T; context?: Context };
```
lastmjs, line 21: passes should be succeed, also add an error message thing

test/index.ts
```diff
@@ -1,26 +1,37 @@
...
     Err: E;
 }>;
 
-export type Ok<T> = {
-    Ok: T;
+// type AzleTestContext = any; // TODO I think I want this to be object eventually
```
lastmjs, line 25: Gone

lastmjs (changes requested):

examples/basic_bitcoin/test/tests.ts
bdemann: done

examples/apollo_server/test/tests.ts
bdemann: done

examples/candid_encoding/test/tests.ts
bdemann: done I did succeeds instead because a boolean value should be named as a true false question and not a verb

bdemann: well I guess in that case isSuccessful would be more correct

bdemann: Although looking at it closer its not always a boolean, its a generic T that just happens to be a boolean in every case

examples/candid_keywords/test/tests.ts
bdemann: done

examples/complex_types/test/tests.ts
bdemann: done

examples/ethers_base/test/tests.ts
bdemann: done

test/index.ts
bdemann: done

test/index.ts
bdemann: done

examples/candid_encoding/test/tests.ts
bdemann: So we get rid of the generic and rename it to isSuccessful

examples/audio_recorder/test/tests.ts
```diff
@@ -1,155 +1,115 @@
...
-
-                return {
-                    Ok:
-                        user.username === 'lastmjs' &&
-                        user.recordingIds.length === 0
-                };
+                const expectedUsername = 'lastmjs';
+                const expectedRecordingCount = 0;
+
+                return createTestResult<Context>(
```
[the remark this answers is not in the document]
bdemann: Easy, just shove all of the values into a tuple
```javascript
return equals<Context>(
    [user.username, user.recordingIds.length],
    [expectedUsername, expectedRecordingCount],
    { context: { ...context, user } }
);
```
This will maintain the majority of the information we got from the custom error message (and all of the information that is being tests) and it doesn't require any changes to any code. 

Anything that doesn't follow this patter shouldn't be using an equal method anyways.

And equals should be renamed to testEquality. 

Oh yes, and then we can get rid of that customEqual business and rename that to just equals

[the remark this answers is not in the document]
bdemann: Here are some failed attempts for posterity
```javascript
type Equal = {
    actual: any;
    expected: any;
    func?: (actual: any, expected: any) => boolean;
};

type MultiEqualsOptions<Context> = {
    failMessage?: string;
    toString?: (value: any) => string;
    context?: Context;
};

export function multiEquals<Context>(
    tests: Equal[],
    options?: MultiEqualsOptions<Context>
): AzleResult<string, Context> {
    const valueToString = options?.toString ?? jsonStringify;
    for (const { func, actual, expected } of tests) {
        const equals = func ?? deepEqual;
        if (equals(actual, expected)) {
            continue;
        } else {
            const message =
                options?.failMessage ??
                `Expected: ${valueToString(
                    expected
                )}, Received: ${valueToString(actual)}`;
            return fail(message);
        }
    }
    return succeed(options?.context);
}

export function otherMultiEquals<A, E, Context>(
    actual: A | any[],
    expected: E | any[],
    options?: EqualsOptions<Context> | EqualsOptions<Context>[]
): AzleResult<string, Context> {
    const equals = options?.customEquals ?? deepEqual;
    const valueToString = options?.toString ?? jsonStringify;

    if (equals(actual, expected)) {
        return succeed(options?.context);
    } else {
        const message =
            options?.errMessage ??
            `Expected: ${valueToString(expected)}, Received: ${valueToString(
                actual
            )}`;
        return fail(message);
    }
}
```

bdemann:
closing this in favor of using jest, we will keep the branch around until jest is fully implemented though because it will have some helpful updates to the tests

PR #1828, opened 2024-06-12
Update all examples that start with "b" to use jest

bdemann:


.github/workflows/test.yml
```diff
@@ -321,7 +321,7 @@ jobs:
             - if: ${{ needs.release-candidate-deploy.outputs.should_run_tests && !contains(github.head_ref, 'release--') && !(github.ref == 'refs/heads/main' && contains(github.event.head_commit.message, 'Merge pull request') && contains(github.event.head_commit.message, 'demergent-labs/release--')) }}
               shell: bash -l {0}
               working-directory: ${{ matrix.example_directories }}
-              run: AZLE_PROPTEST_NUM_RUNS=5 AZLE_PROPTEST_VERBOSE=true npm test
+              run: AZLE_PROPTEST_NUM_RUNS=5 AZLE_PROPTEST_VERBOSE=true AZLE_INTEGRATION_TEST_LINK_AZLE=true npm test
```
bdemann, line 324: @lastmjs we should discuss this as part of the review

examples/bytes/test/tests.ts
```diff
@@ -1,80 +1,28 @@
...
-            test: async () => {
+export function getTests(bytes_canister: ActorSubclass<_SERVICE>): Test {
+    return () => {
+        describe.each([
+            'example_1_kb.txt',
+            'example_10_kb.txt',
+            'example_100_kb.txt',
+            'example_1000_kb.txt',
+            'example_2000_kb.txt'
+        ])('', (fileName) => {
```
bdemann, line 16: we need to add a description here

.github/workflows/test.yml
bdemann: "I'm not sure you can just do this generally, we need to handle the npm and repo tests. There are tests that run with Azle installed from npm during release, not from the repo, those must be honored correctly. We should discuss." - @lastmjs

examples/bytes/test/tests.ts
bdemann: done

.github/workflows/test.yml
lastmjs: We are testing this now

lastmjs (changes requested):

lastmjs (approved):

.github/workflows/test.yml
lastmjs: We will remove these logs in a subsequent pr, testing in a release ci/cd

lastmjs (approved):

PR #1829, opened 2024-06-12
Update first half of examples that start with "c" to use jest

bdemann:


lastmjs (dismissed):

lastmjs:
Look into previous failures on fs and others, strange npm install issues

PR #1831, opened 2024-06-12
Update examples that start with "d" or "e" to use jest

bdemann:


examples/ethers_base/test/tests.ts
```diff
@@ -7,176 +7,138 @@
...
-        }
-    ];
+                }
+            );
+            const responseText = await response.text();
+
+            expect(responseText).toMatch('transaction sent with hash:');
+        });
+
+        wait('for funds to be available', 30_000);
```
bdemann, line 88: This is weird that this needed to be added... any ideas what's going on here?

examples/ethers_base/test/tests.ts
```diff
@@ -4,179 +4,138 @@
...
-                    Ok: BigInt(responseJson.__bigint__) < 100_000_000_000_000n
-                };
-            }
-        },
-        {
-            name: '/address-balance caller',
-            test: async () => {
-                const response = await fetch(`${origin}/address-balance`, {
+    return () => {
+        it("gets the caller's address using ethers and ThresholdWallet", async () => {
```
lastmjs, line 17: We should change these to the correct HTTP methods and use the force query/update headers

examples/ethers_base/test/tests.ts
lastmjs: Let's take a look

examples/ethers_base/test/tests.ts
```diff
@@ -4,179 +4,138 @@
...
+                method: 'POST',
+                headers: [['Content-Type', 'application/json']],
+                body: JSON.stringify({
+                    address: canisterAddress
+                })
+            });
+            const responseJson = jsonParse(await response.text());
+
+            expect(responseJson).toBeLessThan(100_000_000_000_000n);
+            expect(responseJson).toBeGreaterThan(0n);
```
lastmjs, line 125: Greater than 0?

lastmjs (changes requested):

examples/ethers_base/test/tests.ts
bdemann: done

examples/ethers_base/test/tests.ts
bdemann: removed

examples/ethers_base/test/tests.ts
bdemann: we looked at it and decided to add a 10 second wait after each transfer to be sure that it will work on whatever computer. It is interesting to note that these same tests were failing on my computer even on the main branch

lastmjs (approved):

PR #1832, opened 2024-06-12
Update examples that start with "f" or "g" to use jest

bdemann:


examples/generics/test/test.ts
```diff
@@ -1,13 +1,16 @@
 import { getCanisterId } from 'azle/dfx';
-import { runTests } from 'azle/test';
+import { runTests } from 'azle/test/jest';
 
 import { createActor } from './dfx_generated/generics';
 import { getTests } from './tests';
 
-const genericsCanister = createActor(getCanisterId('generics'), {
+// TODO is it time to get rid of this example?
```
bdemann, line 7: @lastmjs is it?

examples/fetch_ic/test/tests.ts
```diff
@@ -6,281 +6,221 @@ import * as dns from 'node:dns';
...
+            mainPage = pages[1];
+        });
 
-                await mainPage.waitForNetworkIdle();
+        wait('for identity to be set', 5_000);
 
-                const success = await page.evaluate(
-                    () => (window as any).headersObjectSuccess
-                );
+        it('Headers Array', async () => {
```
bdemann, line 43: @lastmjs I'll need some help with the names for these test

examples/file_protocol/test/tests.ts
```diff
@@ -1,99 +1,56 @@
...
-        {
-            name: '/read-test0',
-            test: async () => {
-                try {
-                    const response = await fetch(
-                        `${origin}/read-test0?filename=${'assets/test0.txt'}`
-                    );
-                    const responseText = await response.text();
+    return () => {
+        it('/reads a file in the top level directory', async () => {
```
bdemann, line 10: 'reads a file in the top level directory'

examples/generics/test/tests.ts
```diff
@@ -1,347 +1,245 @@
...
-                        'Arm3' in result &&
-                        Array.isArray(result.Arm3) &&
-                        result.Arm3[0] === 'It did work'
-                };
-            }
-        }
-    ];
+export function getTests(genericsCanister: ActorSubclass<_SERVICE>): Test {
+    return () => {
+        it('simpleResult', async () => {
```
bdemann, line 8: If we decide to move forward with generic do we want to take the time to try and update it a bit and actually have it run in out tests suite? As part of that do we want to make these names more descriptive?

examples/guard_functions/test/tests.ts
```diff
@@ -1,149 +1,89 @@
...
-                        )
-                    };
-                }
-            }
-        }
-    ];
+): Test {
+    return () => {
+        it('successfully calls an unguarded method', async () => {
+            const result = await guardFunctionsCanister.identifierAnnotation();
```
bdemann, line 11: Were these from when we had a different API for guard methods? I don't exactly understand what they are doing

examples/generics/test/test.ts
lastmjs: Yes it is

examples/fetch_ic/test/tests.ts
lastmjs: Yes we need to work on the names now based on the frontend tests

lastmjs (changes requested):

PR #1841, opened 2024-06-17
Update examples that start with "i" to use jest

bdemann:


examples/ckbtc/test/tests.ts
```diff
@@ -63,26 +63,17 @@ export function getTests(): Test {
...
-            if ('Ok' in updateBalanceResult) {
-                throw new Error(
-                    `Expected principal ${config.caller} to not have new UTXOs`
-                );
-            }
-            const errorType = Object.keys(updateBalanceResult.Err)[0];
-
-            expect(errorType).toBe('NoNewUtxos');
+            expect(updateBalanceResult).toEqual(
+                expect.objectContaining({ Err: 'NoNewUtxos' })
```
lastmjs, line 75: We need to look into the shape of this variant, and look into other of these weird cases to make sure we have things correct

lastmjs (changes requested):

lastmjs (approved):

PR #1847, opened 2024-06-19
Class syntax pre and post upgrade and inspect message

lastmjs:


src/lib/stable/execute_method.ts
```diff
@@ -24,11 +24,7 @@ export function executeMethod(
 
     const result = getResult(decodedArgs, callback);
 
-    if (
-        mode === 'init' ||
-        mode === 'postUpgrade' ||
-        mode === 'inspectMessage'
-    ) {
+    if (mode === 'init' || mode === 'postUpgrade') {
```
bdemann, line 27: why did this happen?

tests/end_to_end/candid_rpc/class_syntax/pre_and_post_upgrade/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        entries = stableEntries.reduce((result, entry) => {
+            return {
+                ...result,
+                [entry.key]: entry.value
+            };
+        }, {});
+    }
+
+    @preUpgrade
```
bdemann, line 51: is this not a function call like other decorators?

bdemann (approved):

PR #1850, opened 2024-06-20
implement cross_canister_calls example, notify, trap

lastmjs:


src/lib/stable/ic_apis/notify.ts
```diff
@@ -0,0 +1,42 @@
...
+export function notify(
+    canisterId: Principal | string,
+    method: string,
+    options?: {
+        paramIdls?: IDL.Type[];
+        args?: any[];
+        payment?: bigint;
+    }
+): void {
+    if (globalThis._azleIc === undefined) {
```
bdemann, line 20: Is there any way we could just define it as a javascript function in rust in the first place, the same way as we make the _azleIc ojbect, but just as functions?

bdemann (dismissed):

bdemann (approved):

bdemann (changes requested):
The failing tests should be easy to fix, it looks like you still just have an import for the ic object that is tripping you up

src/lib/stable/ic_apis/notify.ts
lastmjs: We have created an issue to to consider these kinds of broader design concerns before 1.0: https://github.com/demergent-labs/azle/issues/1862

PR #1856, opened 2024-06-24
Call raw

lastmjs:


examples/call_raw/test/tests.ts
```diff
@@ -15,12 +15,7 @@ export function getTests(call_raw_canister: ActorSubclass<_SERVICE>): Test {
                 0n
             );
 
-            if ('Err' in result) {
-                throw new Error(result.Err);
-            }
-
-            expect(result).toHaveProperty('Ok');
-            expect(result.Ok).toMatch('blob');
+            expect(result).toMatch('blob');
```
bdemann, line 18: What kind of errors might we get from call raw? Should we add some tests to make sure those error throw correctly?

src/lib/stable/ic_apis/call.ts
```diff
@@ -23,12 +24,15 @@ export async function call(
...
+        const raw = options?.raw;
 
         // TODO perhaps we should be more robust
         // TODO for example, we can keep the time with these
         // TODO if they are over a certain amount old we can delete them
         globalThis._azleResolveIds[globalResolveId] = (result: ArrayBuffer) => {
-            if (returnIdl === undefined) {
+            if (raw !== undefined) {
+                resolve(new Uint8Array(result));
+            } else if (returnIdl === undefined) {
```
bdemann, line 35: I'm just thinking about this a little bit... in the case where the developer leaves off the returnIDL his function is going to be returning undefined. 

But if he had put the wrong IDL then he would have gotten an error that is something like "failed to serialize <value> to <type>. 

I'm wondering if that would be a more instructive error to help them know that they are missing the return IDL instead of just returning undefined.

examples/call_raw/test/tests.ts
lastmjs: Issue created to address this more broadly: https://github.com/demergent-labs/azle/issues/1863

src/lib/stable/ic_apis/call.ts
lastmjs: This seems like a good idea, let's implement it later and make sure to test it: https://github.com/demergent-labs/azle/issues/1864

PR #1907, opened 2024-07-16
Candid generation

bdemann:


tests/end_to_end/candid_rpc/class_syntax/list_of_lists/src/index.ts
```diff
@@ -153,7 +153,7 @@ export default class {
 
     @query(
         [IDL.Vec(IDL.Vec(IDL.Vec(IDL.Int64)))],
-        IDL.Vec(IDL.Vec(IDL.Vec(IDL.Nat64)))
+        IDL.Vec(IDL.Vec(IDL.Vec(IDL.Int64)))
```
bdemann, line 156: This is interesting, I'm guessing it's because every time it gets serded it will end up a bigint so we never ran into problems?

lastmjs: Our hypothesis is that because of bigints in js, this didn't end up being an actual issue.

src/lib/globals.ts
```diff
@@ -64,6 +65,10 @@ declare global {
     var _azleCanisterMethods: CanisterMethods;
     // eslint-disable-next-line no-var
     var _azleCanisterMethodsIndex: number;
+    // eslint-disable-next-line no-var
+    var _azleCanisterMethodIdlTypes: { [key: string]: IDL.FuncClass };
+    // eslint-disable-next-line no-var
+    var _azleSystemCanisterMethodIdlTypes: IDL.FuncClass[];
```
lastmjs, line 71: Change to `_azleInitAndPostUpgradeIdlTypes`

src/lib/globals.ts
```diff
@@ -79,6 +84,10 @@ if (globalThis._azleInsideCanister) {
         callbacks: {}
     };
 
+    globalThis._azleCanisterMethodIdlTypes = {};
```
lastmjs, line 87: Did you initialize these in stable global?

src/lib/stable/init.ts
```diff
@@ -16,6 +16,11 @@ export function init<This, Args extends any[], Return>(
             index
         };
 
+        globalThis._azleSystemCanisterMethodIdlTypes = [
```
lastmjs, line 19: Just push

src/lib/stable/post_upgrade.ts
```diff
@@ -16,6 +16,11 @@ export function postUpgrade<This, Args extends any[], Return>(
             index
         };
 
+        globalThis._azleSystemCanisterMethodIdlTypes = [
```
lastmjs, line 19: push and rename the global variable

lastmjs (changes requested):

src/lib/globals.ts
bdemann: done

src/lib/stable/init.ts
bdemann: done

src/lib/stable/post_upgrade.ts
bdemann: done

src/lib/globals.ts
bdemann: done

lastmjs (approved):

PR #1970, opened 2024-07-24
Trap property tests

lastmjs:


tsconfig.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "strict": true,
         "target": "ES2020",
+        "module": "ES2020",
```
bdemann, line 5: out of curiosity why do we need this?

test/index.ts
```diff
@@ -115,3 +117,23 @@ function processEnvVars(): {
         shouldRunBenchmarks
     };
 }
+
+export const defaultParams = {
+    numRuns: Number(process.env.AZLE_PROPTEST_NUM_RUNS ?? 1),
+    endOnFailure: process.env.AZLE_PROPTEST_SHRINK === 'true' ? false : true
+};
+
+export async function getCanisterActor<T>(
```
bdemann, line 126: should we provide a way to add an agent?

test/index.ts
```diff
@@ -115,3 +117,23 @@ function processEnvVars(): {
         shouldRunBenchmarks
     };
 }
+
+export const defaultParams = {
```
bdemann, line 121: should this be defaultPropertyTestParams or something to make it clear they are for prop tests?

tests/property/ic_api/trap/src/index.ts
```diff
@@ -0,0 +1,56 @@
+// TODO to improve these tests we would also test init, postUpgrade, and preUpgrade
```
bdemann, line 1: could you briefly explain why you didn't

bdemann (changes requested):
Should I be worried about how many tests are failing?

tests/property/ic_api/trap/src/index.ts
lastmjs: I'll just remove this commented code

test/index.ts
lastmjs: Fixed

tests/property/ic_api/trap/src/index.ts
lastmjs: Removed

bdemann (dismissed):

lastmjs:
Trap has a problem, it isn't terminating. I am looking into it

lastmjs:
I just needed to update a name I had changed in a previous PR

PR #1975, opened 2024-07-29
Time property tests

lastmjs:


tests/property/ic_api/time/src/index.ts
```diff
@@ -0,0 +1,55 @@
+// TODO to improve these tests we would also test init, postUpgrade, and preUpgrade
```
bdemann, line 1: how many of these comments apply to time and how many are just copied from trap?

bdemann: Can you access stable memory from inspectMessage? Could we store the time instead of trapping? Trapping is probably fine, I'm mostly curious

tests/property/ic_api/time/test/tests.ts
```diff
@@ -0,0 +1,126 @@
...
+import fc from 'fast-check';
+
+import { _SERVICE as Actor } from './dfx_generated/canister/canister.did';
+
+export function getTests(): Test {
+    return () => {
+        it('should return time from a query method', async () => {
+            const actor = await getCanisterActor<Actor>('canister');
+
+            await fc.assert(
```
bdemann, line 11: what benefit do we get by using fast check here?

bdemann (changes requested):

tests/property/ic_api/time/src/index.ts
lastmjs: I will prove that you either can or cannot persist state in inspectMessage

lastmjs: Proven

PR #1978, opened 2024-07-30
Add es module 2022

lastmjs:


bdemann (dismissed):
I'm assuming that by the time we are getting ready to merge this one that you will have answered why we need this as part of one of the previous prs

bdemann (dismissed):

PR #1999, opened 2024-08-08
Rust experimental

lastmjs:


bdemann (approved):

bdemann (changes requested):
It looks like any test that has assets has to be experimental since it's going to be using the file uploader.

lastmjs:
I have addressed the issue and added the experimental flag where necessary in the class-based tests

bdemann (approved):

PR #2000, opened 2024-08-08
Refactor rust

lastmjs:


src/compiler/rust/canister/src/lib.rs
```diff
@@ -246,294 +57,24 @@ fn run_event_loop(context: &mut wasmedge_quickjs::Context) {
...
 #[ic_cdk_macros::update(guard = guard_against_non_controllers)]
-fn reload_js(
+fn _azle_reload_js(
     timestamp: u64,
     chunk_number: u64,
     js_bytes: Vec<u8>,
     total_len: u64,
     function_index: i32,
-    experimental: bool,
+    experimental: bool, // TODO we should be able to get rid of this parameter
```
bdemann, line 68: Can we?

src/compiler/rust/canister/src/stable_b_tree_map/mod.rs
bdemann, line 1: Can we move this mod file to src/stable_b_tree_map.rs?

src/compiler/rust/canister/src/autoreload/mod.rs
bdemann, line 1: Can we move this to src/autoreload.rs

bdemann (changes requested):
If you could also take 30 seconds to compile with verbose and clean up some of the rust warnings that would be great, I know that there are a number of really easy ones to clean up, like unnecessary unsafe blocks and their might be some unused imports and variables, especially after a refactor like this, it would be good to make sure there is not accidental dead code left over

src/compiler/rust/canister/src/lib.rs
lastmjs: It has been removed

src/compiler/rust/canister/src/stable_b_tree_map/mod.rs
lastmjs: Done

src/compiler/rust/canister/src/autoreload/mod.rs
lastmjs: Done

bdemann (approved):

PR #2006, opened 2024-08-20
Class based prop tests up to float

bdemann:


property_tests/tests/class_syntax/blob/dfx.json
```diff
@@ -0,0 +1,16 @@
...
+        "canister": {
+            "type": "azle",
+            "main": "src/index.ts",
+            "candid_gen": "automatic",
+            "declarations": {
+                "output": "test/dfx_generated/canister",
+                "node_compatibility": true
+            },
+            "custom": {
+                "experimental": true
```
lastmjs, line 12: Experimental must be removed from all class-based examples if possible

property_tests/tests/class_syntax/blob/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "scripts": {
+        "test": "tsx test/test.ts"
+    },
+    "dependencies": {
+        "azle": "0.23.0"
+    },
+    "devDependencies": {
+        "tsx": "^4.15.7",
+        "typescript": "^5.2.2"
```
lastmjs, line 10: Can you get rid of typescript and see if that works? If it works, can you remove it from all property tests?

lastmjs (changes requested):

property_tests/tests/class_syntax/blob/dfx.json
bdemann: Done

property_tests/tests/class_syntax/blob/package.json
bdemann: done

PR #2008, opened 2024-08-20
Class based prop tests prims

bdemann:


bdemann:
Made this on accident

PR #2009, opened 2024-08-20
Class based prop tests complex types

bdemann:


lastmjs (changes requested):
Remove TypeScript in package.json and experimental in dfx.json if possible

PR #2010, opened 2024-08-20
Class based prop tests canister methods

bdemann:


property_tests/arbitraries/http/request_arb.ts
```diff
@@ -117,16 +126,44 @@ export function HttpRequestArb(): fc.Arbitrary<
...
+                idl: 'HttpRequest'
             }
         };
     });
 }
+
+function generateVariableAliasDeclarations(syntax: Syntax): string {
+    if (syntax === 'class') {
+        return /*TS*/ `
+        export type RequestHeaderField = [string, string];
```
lastmjs, line 149: Flip order please

lastmjs (changes requested):
Remove TypeScript in package.json and experimental in dfx.json if possible

property_tests/arbitraries/http/request_arb.ts
bdemann: done

PR #2013, opened 2024-08-22
massive reorganization and rename

bdemann:
If this passes then it might be good to just merge in, it's just moving files around and updating imports


bdemann:
closing in favor of https://github.com/demergent-labs/azle/pull/2014

PR #2015, opened 2024-08-23
Class based prop tests messured move

bdemann:


lastmjs (changes requested):
Please update with the final changes

lastmjs (approved):

PR #2031, opened 2024-08-29
Update GitHub workflows

bdemann:
- [x] Make sure this stuff works on a release branch
- [x] Make sure bash conditionals are correct and using the correct equality operator
- [x] Update the ids of the final two tests in the yml and update the corresponding ids/names in github
- [x] Make sure that the release tests test everything using the AZLE_TEST_RUN_ON_RELEASE env var
    - [x] class_syntax/new
    - [x] large_files 
        - [x] 250MiB
        - [x] 1 GiB
        - [x] 2 GiB
   - [x] http_server/new
   - [x] Additionally some are filtered entirely if we are on a release
        - [x] examples/basic_bitcoin
        - [x] examples/bitcoin_psbt
        - [x] examples/ckbtc
        - [x] tests/end_to_end/http_server/ethers_base
        - [x] tests/end_to_end/http_server/http_outcall_fetch
        - [x] tests/end_to_end/http_server/ic_evm_rpc
        - [x] tests/property/candid_rpc/class_api/stable_b_tree_map
        - [x] tests/property/candid_rpc/functional_api/stable_b_tree_map
   - [ ] My current thought is that these that are filtered we hope to one day run on regular tests but are too flaky. Meanwhile tests like new will only ever run on releases. So those are proper to have filtered in their package.json to denote the permanentality of their skipping except on release, right?


.github/workflows/test.yml
```diff
@@ -1,323 +1,83 @@
...
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//'%'/'%25'}"
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//$'\n'/'%0A'}"
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//$'\r'/'%0D'}"
-                  echo "::set-output name=example_directories::$EXAMPLE_DIRECTORIES"
-            - id: should_run_tests
+                  directories: |
+                      ./examples
+                      ./tests
+                  exclude_dirs: |
+                      ${{ env.IS_FEATURE == 'true' && format('
```
lastmjs, line 36: `IS_FEATURE` must be renamed to `IS_FEATURE_BRANCH_PR`

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
+name: Deploy release candidate
+on:
+    push:
+        branches:
+            - main
+    pull_request: # Runs on pull requests to any branch
+env:
+    IS_RELEASE_CANDIDATE: ${{ contains(github.head_ref, 'release--') }}
```
lastmjs, line 8: Also we should say `BRANCH_PR` or `BRANCH_MERGE or COMMIT`

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+    IS_FEATURE: ${{ !contains(github.head_ref, 'release--') }}
```
lastmjs, line 9: IS_FEATURE_BRANCH_PR

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+            - uses: actions/checkout@v4
+            - name: Get all test dirs
+              id: get_tests
+              uses: ./.github/actions/generate_tests
+              with:
+                  directories: |
+                      ./examples
+                      ./tests
+
+    determine-is-manual-release:
```
lastmjs, line 29: Rename to `determine-should-release`

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+        if: ${{ contains(github.head_ref, 'release--') }}
+        name: Determine if this job is for a manual release
+        runs-on: ubuntu-latest
+        outputs:
+            is_manual_release: ${{ steps.determine_is_manual_release.outputs.is_manual_release }}
+        steps:
```
lastmjs, line 35: Declaratize this process

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+                      IS_AUTOMATED_RELEASE="true"
+                  fi
+
+                  if [[ "$IS_AUTOMATED_RELEASE" == "false" ]]; then
+                      echo "is_manual_release=false" >> "$GITHUB_OUTPUT"
+                  else
+                      echo "is_manual_release=true" >> "$GITHUB_OUTPUT"
+                  fi
+
+    release-candidate-deploy:
```
lastmjs, line 54: Rename to just `release`

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+            is_manual_release: ${{ steps.determine_is_manual_release.outputs.is_manual_release }}
+        steps:
+            - uses: actions/checkout@v4
+            - id: determine_is_manual_release
+              run: |
+                  BRANCH_NAME="${{ github.head_ref }}"
+                  RELEASE_VERSION="${BRANCH_NAME:9}"
+                  COMMIT_MESSAGE=$(git log -1 --pretty=format:"%s")
+
+                  IS_AUTOMATED_RELEASE="false"
```
lastmjs, line 43: I'm hoping if we rename everything to `should_release` and `release` that this will become more straight-forward

.github/workflows/build.yml
```diff
@@ -0,0 +1,93 @@
...
+                  fi
+
+                  if [[ "$IS_AUTOMATED_RELEASE" == "false" ]]; then
+                      echo "is_manual_release=false" >> "$GITHUB_OUTPUT"
+                  else
+                      echo "is_manual_release=true" >> "$GITHUB_OUTPUT"
+                  fi
+
+    release-candidate-deploy:
+        name: Deploy release candidate for release branches
```
lastmjs, line 55: This name should change accordingly

.github/workflows/test.yml
```diff
@@ -1,323 +1,85 @@
...
+# TODO give every step a name
+
+name: Tests
 on:
     push:
         branches:
             - main
     pull_request: # Runs on pull requests to any branch
+env:
+    IS_RELEASE_CANDIDATE: ${{ contains(github.head_ref, 'release--') }}
```
lastmjs, line 15: Rename variables as discussed in other comments

.github/workflows/test.yml
```diff
@@ -1,323 +1,85 @@
...
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//'%'/'%25'}"
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//$'\n'/'%0A'}"
-                  EXAMPLE_DIRECTORIES="${EXAMPLE_DIRECTORIES//$'\r'/'%0D'}"
-                  echo "::set-output name=example_directories::$EXAMPLE_DIRECTORIES"
-            - id: should_run_tests
+                  directories: |
+                      ./examples
+                      ./tests
+                  exclude_dirs: |
+                      ${{ env.IS_FEATURE == 'true' && format('
```
lastmjs, line 36: Is there no more elegant way to do this than bash? Could we somehow do it with just yaml variables and if statements or something?

.github/workflows/test.yml
```diff
@@ -1,323 +1,85 @@
...
+              # Determine if tests should run for this pr. If it's a manually submitted release candidate no tests will run.
               run: |
                   BRANCH_NAME="${{ github.head_ref }}"
                   RELEASE_VERSION="${BRANCH_NAME:9}"
                   COMMIT_MESSAGE=$(git log -1 --pretty=format:"%s")
-                  if [[ "${{ contains(github.head_ref, 'release--') }}" == "true" && "$COMMIT_MESSAGE" != "azle-bot automated release $RELEASE_VERSION" ]]
-                  then
-                    ./publish-github-action.sh $RELEASE_VERSION ${{ toJSON(steps.example_directories.outputs.example_directories) }}
+
+                  IS_AUTOMATED_RELEASE="false"
```
lastmjs, line 62: This seems repeated...is there no way we can package this up somehow and only run it once?

.github/workflows/test.yml
```diff
@@ -1,323 +1,85 @@
...
+                      echo "should_run_tests=false" >> "$GITHUB_OUTPUT"
                   else
-                    echo "::set-output name=should_run_tests::true"
+                      echo "should_run_tests=true" >> "$GITHUB_OUTPUT"
                   fi
-    basic-integration-tests:
-        needs: release-candidate-deploy
+
+    run-test:
+        # Tests can either run against the current code of Azle found in the repository, or the code deployed by the GitHub Action to npm.
```
lastmjs, line 74: Can you redo the comments everywhere? Seems like maybe these could be updated, and we should use our variables etc to explain things better

.github/workflows/test.yml
```diff
@@ -1,323 +1,85 @@
...
+        # Feature branch pull requests (pull requests without release-- in the base branch name) will run all tests against the code found in the repository.
+        # Release branch pull requests (pull requests with release-- in the base branch name) will run all tests against the code found in the repository and the code deployed by the GitHub Action to npm.
+        # Pushes to main will run all tests against the code in the repository if the latest commit was not a merge of a release branch, and will run tests against the code in the repository and the code deployed by the GitHub Action to npm otherwise.
+        name: '${{matrix.tests.name}} | ${{matrix.tests.type}} | ${{matrix.tests.syntax}} | ${{matrix.tests.api}} | ${{matrix.azle_source}}'
+        needs:
+            - determine-should-run-tests
```
lastmjs, line 80: Can't should run tests be determined by the build workflow? Can we connect them somehow? Can we determine should release and should run tests at the same time?

.github/workflows/test.yml
```diff
@@ -340,70 +102,120 @@ jobs:
...
-              run: npx azle install-dfx-extension
-            - if: ${{ needs.release-candidate-deploy.outputs.should_run_tests }}
-              working-directory: ${{ matrix.example_directories }}
+
+            - name: Start dfx
+              if: ${{ !env.IS_RELEASE_CANDIDATE }}
+              working-directory: ${{ matrix.tests.path }}
+              run: dfx start --clean --background --host 127.0.0.1:8000 --artificial-delay 0
+
+            - name: Start dfx with artifical delay
```
lastmjs, line 155: artificial

.github/workflows/test.yml
```diff
@@ -340,70 +102,120 @@ jobs:
...
+                  fi
+
+                  # Check if it's a merge from a release candidate into main
+                  if [[ ${{ github.ref }} == "$MAIN_BRANCH" && $COMMIT_MESSAGE == *"$RELEASE_PR_MERGE_MSG"* && $COMMIT_MESSAGE == *"$RELEASE_REPO_MSG"* ]]; then
+                      IS_MERGE_TO_MAIN_FROM_RELEASE=true
+                  else
+                      IS_MERGE_TO_MAIN_FROM_RELEASE=false
+                  fi
+
+                  # Determine the number of runs based on the conditions
```
lastmjs, line 194: Is this the best place for the variables to live?

.github/workflows/test.yml
```diff
@@ -340,70 +102,120 @@ jobs:
...
+                  AZLE_PROPTEST_NUM_RUNS=$RUNS AZLE_PROPTEST_VERBOSE=true npm test
               shell: bash -l {0}
-              working-directory: ${{ matrix.example_directories }}
-              run: AZLE_PROPTEST_NUM_RUNS=100 AZLE_PROPTEST_VERBOSE=true npm test
+              working-directory: ${{ matrix.tests.path }}
 
-    check-basic-integration-tests-success:
-        needs: basic-integration-tests
+    # These final jobs are designed to ensure that all jobs spun up from the matrix in the run-test have succeeded
+    check-basic-integration-test-success:
```
lastmjs, line 208: Should we change these to `check-test-success` or something else? I know I will have to change them in the repo settings.

.github/actions/generate_tests/action.yml
```diff
@@ -0,0 +1,29 @@
+name: Generate Tests
```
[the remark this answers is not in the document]
bdemann: done

.github/actions/generate_tests/action.yml
```diff
@@ -0,0 +1,29 @@
...
+description: Generates a list of directories of npm projects with a tests script
```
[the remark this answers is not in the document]
bdemann: done

.github/actions/generate_tests/action.yml
```diff
@@ -0,0 +1,29 @@
...
+inputs:
+    directories:
+        description: List of directories to search for npm projects with tests.
```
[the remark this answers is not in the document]
bdemann: done

.github/actions/generate_tests/action.yml
```diff
@@ -0,0 +1,29 @@
+name: Generate Tests
+description: Generates a list of directories of npm projects with an npm test script
+inputs:
+    directories:
+        description: List of directories to search for npm projects with tests
```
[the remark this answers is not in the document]
bdemann: done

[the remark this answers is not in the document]
bdemann: done

.github/actions/generate_tests/action.yml
```diff
@@ -0,0 +1,29 @@
...
+        required: true
+    exclude_dirs:
+        description: List of directories to exclude from the search
+        required: false
+        default: ''
+outputs:
+    all_tests:
```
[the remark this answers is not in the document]
bdemann: done

.github/scripts/generate_tests.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# Inputs from action.yml
+BASE_DIR="."
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
+name: Deploy release candidate
+on:
+    push:
+        branches:
+            - main
+    pull_request: # Runs on pull requests to any branch
+env:
+    IS_RELEASE_CANDIDATE: ${{ contains(github.head_ref, 'release--') }}
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+    IS_FEATURE: ${{ !contains(github.head_ref, 'release--') }}
+    DFX_VERSION: 0.21.0
+    NODE_VERSION: 20
+jobs:
+    get-all-test-dirs:
+        if: ${{ contains(github.head_ref, 'release--') }}
+        name: Get all test dirs
+        runs-on: ubuntu-latest
+        outputs:
+            all_tests: ${{ steps.get_all_tests.outputs.all_tests }}
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+        steps:
+            - uses: actions/checkout@v4
+            - name: Get all test dirs
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/build.yml
bdemann: done

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+    NODE_VERSION: 20
+jobs:
+    get-all-test-dirs:
+        if: ${{ contains(github.head_ref, 'release--') }}
+        name: Get all test dirs
+        runs-on: ubuntu-latest
+        outputs:
+            all_tests: ${{ steps.get_all_tests.outputs.all_tests }}
+        steps:
+            - uses: actions/checkout@v4
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/build.yml
bdemann: done

.github/workflows/build.yml
bdemann: done

.github/workflows/build.yml
bdemann: done

.github/workflows/test.yml
bdemann: done

.github/workflows/test.yml
bdemann: I've heard it both ways

.github/workflows/test.yml
bdemann: yes, that task is already mentioned above, I was holding off until last minute because it was making my tests fail since the repo settings weren't updated yet

.github/workflows/test.yml
bdemann: yeah we could put it into an action

done

.github/workflows/test.yml
bdemann: moved it into it's own action. I don't think there is any reason to prevent it from running twice, it's not that long of a task, and I'm not even sure it's possible for workflows to communicate with each other. But as an action the code isn't duplicated at least

.github/workflows/test.yml
bdemann: it's gone I think the other comments are sufficient

.github/workflows/build.yml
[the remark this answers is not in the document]
bdemann: I don't think so. There are a number of factors that go into what variable we can have here. 1 the logic has to be simple enough to have a one liner. 2 there are only so many things we can check at this high a level. 3 there aren't that many variable that are used in more than one place. etc. I think the number of variables we have right now is good

.github/workflows/build.yml
bdemann: done

.github/workflows/test.yml
bdemann: The current approach uses inline Bash to dynamically set exclude_dirs, which I believe keeps the logic compact and flexible within a single job. Alternatives like splitting into multiple jobs or using external files could be considered, but they might add unnecessary complexity. However, I'm open to refining this if you feel strongly about it!

.github/workflows/test.yml
bdemann: I think so, I consolidated one into IS_MAIN_BRANCH and moved it up with the others because it could just in case, maybe it's not too much future proofing... and the I obliterated as many of them as I could. I think the result is more declarative.

.github/workflows/build.yml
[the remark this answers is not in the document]
bdemann: I did add IS_MAIN

.github/workflows/build.yml
bdemann: I think it is

.github/scripts/publish_github_action.sh
```diff
@@ -8,13 +8,17 @@ VERSION=$1
...
+directories=$(echo "$directories_json_string" | jq -c -r '.[] | .path')
 
 sed -E -i "s/(\"version\": \")(.*)(\")/\1$VERSION\3/" package.json
 sed -E -i "s/(\"version\": \")(.*)(\")/\1$VERSION\3/" dfx_extension/extension.json
 # TODO we need to keep the dependencies.json file up-to-date as well
 npm install
 
+# Build the binary templates
+npx azle template
+npx azle template --expirimental
```
lastmjs, line 20: This is spelled wrong

.github/actions/get_tests/generate_json.js
lastmjs, line 1: Redo this immutably and nicely: no mutations, no else if, no multiple levels of if statements

.github/actions/should_release/action.yml
```diff
@@ -0,0 +1,29 @@
+# TODO I feel like the wording on this is a little clunky and/or ethereal.
```
lastmjs, line 1: Get rid of these comments

.github/actions/should_release/action.yml
```diff
@@ -0,0 +1,29 @@
...
+
+        - id: determine_should_release
+          run: |
+              BRANCH_NAME="${{ github.head_ref }}"
+              RELEASE_VERSION="${BRANCH_NAME:9}"
+              COMMIT_MESSAGE=$(git log -1 --pretty=format:"%s")
+              IS_RELEASE_BRANCH_PR="${{ contains(github.head_ref, 'release--') }}"
+
+              SHOULD_RELEASE="false"
+              if [[ $IS_RELEASE_BRANCH_PR == "true" && "$COMMIT_MESSAGE" != "azle-bot automated release $RELEASE_VERSION" ]]; then
```
lastmjs, line 24: Create a declarative variable for the commit message text, the name should explain what that text is for

lastmjs: Actually move this message up and pass it as an input to this action and the publish script. Call it: AUTOMATED_RELEASE_COMMIT_MESSAGE

lastmjs: Actually, don't do any of that. Change the if statement to check for an exact message, which will be the name of the release branch (e.g. release--0.25.0-rc.0)

.github/workflows/release.yml
```diff
@@ -1,63 +1,50 @@
...
+        runs-on: ubuntu-latest
+        outputs:
+            should_release: ${{ steps.determine_should_release.outputs.should_release }}
+        steps:
+            - uses: actions/checkout@v4
+
+            - id: determine_should_release
+              uses: ./.github/actions/should_release
+
+    get-tests:
```
lastmjs, line 23: Rename test to testInfo...get-test-infos...etc

.github/workflows/release.yml
```diff
@@ -1,63 +1,50 @@
...
+        needs: determine-should-release
+        if: ${{ contains(github.head_ref, 'release--') && needs.determine-should-release.outputs.should_release }}
```
lastmjs, line 25: Can you give this a name? The release-- thing? Can you reuse the same variable location as in the determine_should-release action?

.github/workflows/release.yml
```diff
@@ -1,63 +1,50 @@
...
-
-    release-candidate-deploy:
-        name: Deploy release candidate for release branches
+    release:
+        name: Deploy release
         # Only run this job if it's a release branch. This job will run instead of run-tests and will automatically publish another commit which will be tested
-        if: ${{ !needs.determine-is-manual-release.outputs.is_manual_release }}
+        if: ${{ needs.determine-should-release.outputs.should_release == 'true' }}
         needs:
             - get-tests
```
lastmjs, line 46: Can we flip the order of these? Seems nice to put them in the order they will run...

.github/workflows/test.yml
```diff
@@ -3,22 +3,35 @@
...
             - main
     pull_request: # Runs on pull requests to any branch
 env:
-    IS_RELEASE_CANDIDATE: ${{ contains(github.head_ref, 'release--') }}
-    IS_FEATURE: ${{ !contains(github.head_ref, 'release--') }}
-    DFX_VERSION: 0.21.0
+    DFX_VERSION: 0.22.0
     NODE_VERSION: 20
+    IS_MAIN_BRANCH: ${{ github.ref == 'refs/heads/main' }}
+    IS_RELEASE_BRANCH_PR: ${{ contains(github.head_ref, 'release--') }}
```
lastmjs, line 16: Is there a startsWith instead of contains? Would be safer

.github/workflows/test.yml
```diff
@@ -148,53 +132,22 @@ jobs:
...
 
             - name: Start dfx
-              if: ${{ !env.IS_RELEASE_CANDIDATE }}
+              if: ${{ env.IS_FEATURE_BRANCH_PR == 'true' }}
               working-directory: ${{ matrix.tests.path }}
               run: dfx start --clean --background --host 127.0.0.1:8000 --artificial-delay 0
 
-            - name: Start dfx with artifical delay
-              if: ${{ env.IS_RELEASE_CANDIDATE }}
+            - name: Start dfx with artifcial delay
```
lastmjs, line 139: artificial is spelled incorrectly, and isn't the name here wrong? Shouldn't it be swapped?

.github/workflows/test.yml
```diff
@@ -148,53 +132,22 @@ jobs:
...
-                      IS_MERGE_TO_MAIN_FROM_RELEASE=false
-                  fi
+                  IS_MERGE_TO_MAIN_FROM_RELEASE=${{ env.IS_MAIN_BRANCH == 'true' && contains(github.event.head_commit.message, 'Merge pull request') && contains(github.event.head_commit.message, 'demergent-labs/release--') }}
 
-                  # Determine the number of runs based on the conditions
-                  if [[ $IS_MERGE_TO_MAIN_FROM_RELEASE == true ]]; then
+                  if $IS_MERGE_TO_MAIN_FROM_RELEASE; then
                       RUNS=100
-                  elif [[ $IS_RELEASE_PR == true ]]; then
+                  elif ${{ env.IS_RELEASE_BRANCH_PR == 'true' && env.IS_MAIN_BRANCH == 'false' }}; then
```
lastmjs, line 150: Can you give a descriptive name to each and not use elif? Just use if for both cases. And set runs to 5 before anything else

lastmjs (changes requested):

.github/workflows/test.yml
lastmjs: This is a joke right?

.github/scripts/publish_github_action.sh
bdemann: done

.github/actions/get_tests/generate_json.js
bdemann: done

.github/actions/should_release/action.yml
bdemann: done

.github/workflows/release.yml
bdemann: sure

.github/workflows/test.yml
bdemann: sure

.github/workflows/test.yml
bdemann: looks like there is

.github/workflows/release.yml
bdemann: I don't think it's worth the overhead. When we finalize the review I can show you what it would take

But we would either have to 1. dilute an already pretty concise job with an output that doesn't really fit (and is actually required by the job, or 2. make a brand new job just for this one variable. That job will have to run to completion before any of the other jobs can start. 

I understand that the desire to do this, and I tried many times to make this happen but ultimately decided that it wasn't worth the trouble. It is used very infrequently, and is reasonably straight forward as is

.github/workflows/release.yml
bdemann: done

.github/actions/should_release/action.yml
bdemann: done

.github/actions/get_test_infos/action.yml
```diff
@@ -0,0 +1,46 @@
+name: Get test infos
+description:
+    'Gets a list of test info objects for each npm projects with an npm test script
+    The shape of the object is
+    {
+    path: string, // The path to the test
+    name: string, // The name of the test
+    type: string, // The type of test (e.g. ex (example), prop (property test), e2e, (end to end test) etc)
+    syntax: string, // The syntax of the tests (e.g. http_server or candid_rpc)
```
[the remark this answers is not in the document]
lastmjs: Let's redo the data structure to have just name, path, abbreviated path...and we'll display the name first, the npm/repo, the abbreviated path...something

.github/actions/get_test_infos/generate_json.js
lastmjs, line 1: A lot of this file will change because of the new test info data structure

.github/workflows/test.yml
```diff
@@ -136,21 +136,22 @@ jobs:
               working-directory: ${{ matrix.tests.path }}
               run: dfx start --clean --background --host 127.0.0.1:8000 --artificial-delay 0
 
-            - name: Start dfx with artifcial delay
+            - name: Start dfx with artificial delay
```
lastmjs, line 139: Change names

.github/workflows/release.yml
```diff
@@ -62,6 +62,12 @@ jobs:
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+            - name: Install curl (if necessary)
```
lastmjs, line 65: Get rid of if necessary

.github/scripts/install_global_dependencies.sh
lastmjs, line 1: Let's change the dependency saving in our build process to save to a json file...then in here we can use jq and hopefully simplify the parsing.

lastmjs: Make it so that we have a command like `npx azle install-rust-env` that will read the saved version file and create the Rust environment on any developer machine and on GitHub

lastmjs (changes requested):

.github/workflows/test.yml
bdemann: done

.github/workflows/release.yml
bdemann: done

.github/actions/get_test_infos/generate_json.js
bdemann: done

.github/actions/get_test_infos/action.yml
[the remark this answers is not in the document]
bdemann: done

.github/actions/get_test_infos/dir_to_test_info.js
lastmjs, line 1: Use tsx and .ts

.github/actions/get_test_infos/discover_test_dirs.js
```diff
@@ -0,0 +1,43 @@
+import { promises as fs } from 'fs';
+import path from 'path';
+
+// Discover directories containing a package.json file with a test script
+export async function discoverTestDirs(dirsToSearch) {
+    const directories = [];
```
lastmjs, line 6: Make immutable

.github/actions/get_test_infos/discover_test_dirs.js
```diff
@@ -0,0 +1,43 @@
...
+    await findDirectories(dirsToSearch, directories);
+    return directories;
+}
+
+// Recursively find directories and check for package.json with a test script
+async function findDirectories(currentDir, directories) {
```
lastmjs, line 12: Make all immutable

.github/actions/get_test_infos/get_test_infos.js
```diff
@@ -0,0 +1,22 @@
...
+import { dirToTestInfo } from './dir_to_test_info.js';
+
+export async function getTestInfos(dirs, excludeDirs) {
+    const allDirs = await dirs.reduce(async (accPromise, dir) => {
+        const acc = await accPromise;
+        const discoveredDirs = await discoverTestDirs(dir);
+        return [...acc, ...discoveredDirs];
+    }, Promise.resolve([]));
+
+    const testDirs = [...new Set(allDirs)] // Remove duplicates if any
```
lastmjs, line 11: I don't think we'll have duplicates...

.github/actions/get_test_infos/get_test_infos.sh
```diff
@@ -4,33 +4,12 @@
...
-            if [ -f "$pkg/package.json" ]; then
-                if jq -e ".scripts.test" "$pkg/package.json" > /dev/null; then
-                    echo "$pkg"
-                fi
-            fi
-        done
-    ' sh {} +
-}
+# Convert arrays to comma-separated strings for passing into Node.js
+directories=$(IFS=, ; echo "${DIRECTORIES[*]}")
```
lastmjs, line 8: What is this?

.github/actions/get_test_infos/index.js
```diff
@@ -0,0 +1,22 @@
...
+const [, , dirsArg, excludeDirsArg] = process.argv;
+
+// Convert comma-separated strings to arrays
+const dirs = dirsArg.split(',').filter((path) => path !== '');
+const excludeDirs = excludeDirsArg.split(',').filter((path) => path !== '');
+
+async function main() {
+    try {
+        const result = await getTestInfos(dirs, excludeDirs);
+        console.log(JSON.stringify(result));
```
lastmjs, line 15: Let's write directly to stdout

.github/scripts/install_global_dependencies.sh
```diff
@@ -1,10 +1,16 @@
 #!/bin/bash
 
 # File path for the global dependencies
-GLOBAL_DEPENDENCIES_FILE="$PWD/global_dependencies"
+GLOBAL_DEPENDENCIES_FILE="$PWD/global_dependencies.json"
```
lastmjs, line 4: Let's put all of these dependency things into Azle's root-level package.json file as we discussed.

lastmjs: And let's add node, dfx, etc to try and put all dependency versions into one place

.github/scripts/install_global_dependencies.sh
lastmjs, line 1: Have you made an issue to create a command to allow this to be installed on the dev machine easily?

src/build/stable/utils/log_global_dependencies.ts
```diff
@@ -5,19 +5,67 @@ import { execSyncPretty } from './exec_sync_pretty';
...
+            node: nodeVersion,
+            rustc: rustVersion
+        }
+    };
 
     await outputFile(
-        join(AZLE_PACKAGE_PATH, 'global_dependencies'),
-        globalDependencies
+        join(AZLE_PACKAGE_PATH, 'global_dependencies.json'),
+        JSON.stringify(globalDependencies, null, 2) // Format JSON with 2 spaces indentation
```
lastmjs, line 22: Please format with 4 space indentation...I think that's what we are using in our prettier settings...did you have another reason for this?

src/build/stable/utils/log_global_dependencies.ts
```diff
@@ -5,19 +5,67 @@ import { execSyncPretty } from './exec_sync_pretty';
...
     );
 }
+
+// Function stubs for version info, to be filled in later
```
lastmjs, line 26: Is this comment necessary? Seems old

src/build/stable/utils/log_global_dependencies.ts
```diff
@@ -5,19 +5,67 @@ import { execSyncPretty } from './exec_sync_pretty';
...
+    const match = cargoOutput.match(regex);
+
+    if (match !== null && match.length > 1 && typeof match[1] === 'string') {
+        if (match.length > 2 && typeof match[2] === 'string') {
+            return match[2]; // Return the repository link if available
+        } else {
+            return match[1]; // Return the version number if no link is found
+        }
+    } else {
+        return 'unknown';
```
lastmjs, line 69: This should probably throw instead of returning 'unknown' as a string right?

lastmjs (changes requested):

src/build/stable/utils/log_global_dependencies.ts
bdemann: you're right

src/build/stable/utils/log_global_dependencies.ts
bdemann: it is

.github/actions/get_test_infos/index.js
bdemann: done

.github/actions/get_test_infos/get_test_infos.js
bdemann: done

.github/actions/get_test_infos/get_test_infos.sh
bdemann: It's setting the variable IFS to the value of `,`. IFS stands for Internal Field Separator, by setting it to be a comma when we expand DIRECTORIES it will separate them all with a comma

.github/actions/get_test_infos/discover_test_dirs.js
bdemann: done

.github/actions/get_test_infos/dir_to_test_info.js
bdemann: done

.github/actions/get_test_infos/discover_test_dirs.js
bdemann: done

.github/scripts/install_global_dependencies.sh
bdemann: https://github.com/demergent-labs/azle/issues/2095

.github/scripts/install_global_dependencies.sh
bdemann: https://github.com/demergent-labs/azle/issues/2095

.github/actions/get_test_infos/discover_test_dirs.ts
```diff
@@ -0,0 +1,46 @@
+import { promises as fs } from 'fs';
```
lastmjs: Fix this import

.github/actions/get_test_infos/discover_test_dirs.ts
```diff
@@ -0,0 +1,46 @@
...
+import path from 'path';
```
lastmjs: Destructure?

.github/actions/get_test_infos/discover_test_dirs.ts
```diff
@@ -0,0 +1,46 @@
...
+
+                // Recurse into subdirectory
+                return [
+                    ...acc,
+                    ...(hasTestScript ? [fullPath] : []),
+                    ...(await discoverTestDirs(fullPath))
+                ];
+            }
+
+            return [];
```
lastmjs: return acc

.github/actions/get_test_infos/discover_test_dirs.ts
```diff
@@ -0,0 +1,46 @@
...
+        Promise.resolve([] as string[])
+    );
+}
+
+// Check if package.json exists and contains a test script
+async function checkForTestScript(packageJsonPath: string): Promise<boolean> {
+    try {
+        const packageJson = await fs.readFile(packageJsonPath, 'utf-8');
+        const packageData = JSON.parse(packageJson);
+        return (
```
lastmjs: Optional chaining

.github/actions/get_test_infos/get_test_infos.ts
lastmjs, line 1: Can you add function signatures to everything? Can you make sure our eslint prettier whatever stuff is catching these files?

package.json
```diff
@@ -76,5 +76,10 @@
     "lint-staged": {
         "*.{js,ts}": "eslint --cache --fix",
         "**/*": "prettier --write --ignore-unknown"
+    },
+    "globalDependencies": {
```
lastmjs, line 80: Nest this inside of an `"azle"` property.

lastmjs (changes requested):

lastmjs (approved):

lastmjs (approved):

lastmjs (changes requested):
The algorithm is broken now for some reason, the algorithm to determine the paths

PR #2042, opened 2024-08-30
Update GitHub workflows tests

bdemann:


lastmjs:
I'm leaving this comment here for myself from my issue: Make sure that the release tests test everything using the AZLE_TEST_RUN_ON_RELEASE env var

bdemann:
I only needed this for doing tests for the workflow workflow and now that it is done and merged in we can close this

PR #2066, opened 2024-09-04
reduce the max size of files we test for the file uploader until we c…

bdemann:
…an make huge files more stable


lastmjs (approved):

lastmjs (changes requested):
bitcoin_psbt is failing in a way that I don't think it should

PR #2071, opened 2024-09-05
Release  yml test

bdemann:


bdemann:
This was a test release but didn't work because the version number is an essential part of the release processes

PR #2084, opened 2024-09-11
Burn

lastmjs:


tests/property/ic_api/cycles_burn/test/tests.ts
```diff
@@ -0,0 +1,45 @@
+import { getCanisterId } from 'azle/dfx';
+import { defaultPropTestParams, expect, it, Test } from 'azle/test';
+import fc from 'fast-check';
+
+import { execSyncPretty } from '../../../../../src/build/stable/utils/exec_sync_pretty';
```
bdemann, line 5: I think that since this is in a test it should be azle/whatever/exec_sync_prety instead of a relative path

tests/property/ic_api/cycles_burn/test/tests.ts
```diff
@@ -0,0 +1,45 @@
...
+import { createActor } from './dfx_generated/canister';
+
+export function getTests(): Test {
+    return () => {
+        it('should burn the cycles', async () => {
```
bdemann, line 10: I'm wondering if a more descriptive name would be helpful.
like "the canister balance should reflect the amount of cycles burned in the burn command"
That one is very wordy and doesn't fit right in the `it` scheme, but it describes how we are testing that the cycles are burned.

PR #2105, opened 2024-09-18
automatically generate the management canister idl objects and types …

lastmjs:
…from the ic.did file


tests/end_to_end/candid_rpc/class_syntax/call_raw/src/index.ts
```diff
@@ -8,7 +8,7 @@ export default class {
         candidArgs: string,
         payment: bigint
     ): Promise<string> {
-        const result = await call(canisterId, method, {
+        const result = await call<undefined>(canisterId, method, {
```
bdemann, line 11: Why is this undefined? I'm sure there is a good reason, it just looks out of the ordinary so I thought I would ask

bdemann (commented):
Before we merge this could you walk me through the process of updating the management canister index.ts? It's just calling the candidc whatever command right? Should that command be documented in our contributing docs? or somewhere?

PR #2136, opened 2024-09-20
Update release process

bdemann:


lastmjs (changes requested):
Update those two documentation files to explain why checkout shouldn't be included in the actions themselves

lastmjs (dismissed):

PR #2152, opened 2024-10-04
Use reusable workflows to break up tests

bdemann:


.github/workflows/run-test.yml
bdemann, line 1: How is this file different from just cutting it from main as it is?

.github/actions/set_run_conditions/action.yml
lastmjs, line 1: No README for this new action?

.github/actions/set_run_conditions/action.yml
```diff
@@ -0,0 +1,33 @@
...
+runs:
+    using: 'composite'
+    steps:
+        - id: set-conditions
+          run: |
+              # Define conditions using shell variables
+              IS_MAIN_PUSH=${{ github.ref == 'refs/heads/main' && !contains(github.event.head_commit.message, 'demergent-labs/release--') }}
+              IS_MAIN_MERGE_RELEASE=${{ github.ref == 'refs/heads/main' && contains(github.event.head_commit.message, 'demergent-labs/release--') }}
+              IS_RELEASE_PR=${{ startsWith(github.head_ref, 'release--') }}
+              IS_FEATURE_PR=${{ !startsWith(github.head_ref, 'release--') && github.ref != 'refs/heads/main' && github.event.pull_request.draft == false }}
```
lastmjs, line 16: Do you need to check `refs/heads/main`?

.github/actions/set_run_conditions/action.yml
```diff
@@ -0,0 +1,33 @@
...
+              IS_DRAFT_PR=${{ !startsWith(github.head_ref, 'release--') && github.ref != 'refs/heads/main' && github.event.pull_request.draft == true }}
```
lastmjs, line 17: Do you need to check `refs/heads/main`

.github/actions/set_run_conditions/action.yml
```diff
@@ -0,0 +1,33 @@
...
+
+              # Create JSON object
+              CONDITIONS=$(cat <<EOF
+              {
+                "is_main_branch_push": $IS_MAIN_PUSH,
+                "is_main_branch_merge_from_release_push": $IS_MAIN_MERGE_RELEASE,
```
lastmjs, line 23: This name seems off to me

.github/actions/set_run_conditions/action.yml
```diff
@@ -0,0 +1,33 @@
...
+                "is_release_branch_pr": $IS_RELEASE_PR,
+                "is_feature_branch_pr": $IS_FEATURE_PR,
+                "is_feature_branch_draft_pr": $IS_DRAFT_PR
```
lastmjs, line 26: Should we rename this to `is_feature_branch_pr_draft`?

lastmjs (changes requested):

.github/workflows/run_test.yml
```diff
@@ -0,0 +1,150 @@
...
+                      sudo networksetup -setdnsservers Ethernet 9.9.9.9
+                  fi
+
+                  npm install
+
+                  if [[ "${{ matrix.azle_source }}" == "repo" ]]; then
+                      npm link
+                  fi
+
+                  npm run lint
```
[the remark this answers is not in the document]
bdemann: I don't know, it was here when I got here

.github/workflows/run_test.yml
```diff
@@ -0,0 +1,150 @@
...
+            AZLE_IS_FEATURE_BRANCH_DRAFT_PR: ''
+
+        strategy:
+            fail-fast: false # We want to see which example tests succeed and which ones fail, we don't want one example test to cancel the rest
+            matrix: # spins up one job per combination of test and code source (repo or npm).
+                # os: [macos-latest]
+                os: [ubuntu-latest]
+                include_npm:
+                    # Only include npm in the matrix if you've pushed to main and the last commit was a merge of a release branch, or the base branch of the pull request is a release branch
+                    - ${{ (github.ref == 'refs/heads/main' && contains(github.event.head_commit.message, 'Merge pull request') && contains(github.event.head_commit.message, 'demergent-labs/release--')) || contains(github.head_ref, 'release--') }}
```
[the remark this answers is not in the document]
bdemann: I don't think so. Did you have somewhere else in mind?

.github/workflows/run_test.yml
```diff
@@ -0,0 +1,150 @@
...
+            - name: Set condition environment variables
+              run: |
+                  CONDITIONS=$(echo "${{ steps.set-conditions.outputs.conditions }}" | base64 -d)
+                  echo "AZLE_IS_MAIN_BRANCH_PUSH=$(echo $CONDITIONS | jq -r .is_main_branch_push)" >> $GITHUB_ENV
+                  echo "AZLE_IS_MAIN_BRANCH_MERGE_FROM_RELEASE_PUSH=$(echo $CONDITIONS | jq -r .is_main_branch_merge_from_release_push)" >> $GITHUB_ENV
+                  echo "AZLE_IS_RELEASE_BRANCH_PR=$(echo $CONDITIONS | jq -r .is_release_branch_pr)" >> $GITHUB_ENV
+                  echo "AZLE_IS_FEATURE_BRANCH_PR=$(echo $CONDITIONS | jq -r .is_feature_branch_pr)" >> $GITHUB_ENV
+                  echo "AZLE_IS_FEATURE_BRANCH_DRAFT_PR=$(echo $CONDITIONS | jq -r .is_feature_branch_draft_pr)" >> $GITHUB_ENV
+
+            - id: check-conditions
```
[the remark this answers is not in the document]
bdemann: That was what I had in mind

.github/workflows/run_test.yml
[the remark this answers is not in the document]
bdemann: I disagree, let's talk about it

.github/workflows/run_test.yml
```diff
@@ -0,0 +1,150 @@
...
+                tests: ${{ fromJSON(inputs.test_infos) }}
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: ./.github/actions/set_run_conditions
+              id: set-conditions
+
+            - name: Set condition environment variables
+              run: |
+                  CONDITIONS=$(echo "${{ steps.set-conditions.outputs.conditions }}" | base64 -d)
```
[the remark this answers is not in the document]
bdemann: Good idea!

.github/actions/set_run_conditions/action.yml
bdemann: I'll make one

.github/actions/set_run_conditions/action.yml
bdemann: No? I don't think so, but I'm not married to it

.github/actions/set_run_conditions/action.yml
bdemann: maybe not, it's what was here when I got here. I mean we don't want this to be true if the branch is main right?

.github/actions/set_run_conditions/action.yml
bdemann: maybe not, it's what was here when I got here. I mean we don't want this to be true if the branch is main right?

.github/actions/set_run_conditions/action.yml
bdemann: In what way?

.github/workflows/test.yml
```diff
@@ -28,25 +23,29 @@ jobs:
...
-            - id: get-node-version
-              uses: ./.github/actions/get_node_version
+            - id: set-conditions
+              uses: ./.github/actions/set_run_conditions
 
-            - name: Set exclude dirs
-              id: set-exclude-dirs
+            - id: set-exclude-dirs
               run: |
+                  CONDITIONS=$(echo "${{ steps.set-conditions.outputs.conditions }}" | base64 -d)
```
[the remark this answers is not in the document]
bdemann: Yeah, we could just output each of the values instead of making a single object with each one in it

.github/workflows/get_and_run_tests.yml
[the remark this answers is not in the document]
bdemann: yes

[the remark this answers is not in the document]
bdemann: We could possibly move the contents of the run_test.yml file into here I thought about it, but my AI overlords pointed out that the separation was actually pretty nice. What with separation of responsibilities, or something like that. And to be fair, you already criticized the run_test.yml file of being too long and hard to read. Moving it in here would exacerbate that problem, real or not.

Either way this for sure needs to be it's own thing because it drastically simplifies the test.yml. And we never call one of these without the other, in the testing context.

.github/actions/set_run_conditions/action.yml
lastmjs: `is_main_branch_push_from_release_merge` let's rename these ones

.github/workflows/get_and_run_tests.yml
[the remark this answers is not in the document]
lastmjs: You said you would look into this

lastmjs (changes requested):
Just requesting changes on the remaining comments

.github/workflows/run_test.yml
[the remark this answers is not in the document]
bdemann: hopefully you like that one alright

lastmjs (changes requested):

lastmjs (approved):

PR #2153, opened 2024-10-08
Make sure we still can run all of our tests at once

bdemann:


bdemann:
This was only opened to confirm that our runners aren't working as expected

PR #2161, opened 2024-10-14
Parallel release

bdemann:
- [x] try to act on lists of branches instead of single branches when we are doing hundreds of git actions
~- [ ] Make a master workflow that is the only one that runs on push and pull request and determines which workflows should run~
~- [ ] See if we can have only one determine should tests run to launch them~
- [x] tests aren't running on npm code
- [x] Remove testing code
    - [x] make sure that templates are built again
    - [x] restore excluded dirs during release


.github/scripts/publish_github_action.sh
```diff
@@ -26,34 +26,34 @@ else
     npm publish
 fi
 
-# TODO loop through checking for the status instead of sleeping
-echo -e "sleeping for 30 seconds to ensure azle@$VERSION is fully registered on npm"
+# # TODO loop through checking for the status instead of sleeping
```
lastmjs, line 29: Should we delete this comment now? Why keep it?

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
     push:
         branches:
             - main
-    pull_request: # Runs on pull requests to any branch
+    pull_request:
```
lastmjs, line 12: Comment not necessary?

package.json
```diff
@@ -5,7 +5,7 @@
     "scripts": {
         "typecheck": "tsc --noEmit",
         "lint": "if [ \"$npm_config_fix\" ]; then eslint . --ext .js,.ts --cache --fix; else eslint . --ext .js,.ts --cache; fi",
-        "prepare": "husky install",
+        "prepare": "[ -f .git/hooks ] && husky install || true",
```
lastmjs, line 8: Can you explain?

.github/workflows/release.yml
```diff
@@ -43,58 +43,14 @@ jobs:
...
         needs:
             - determine-should-release
             - get-test-infos
-        runs-on: ubuntu-latest
-        env:
+
+        uses: ./.github/workflows/release_parallel.yml
+        secrets:
             GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }} # All commits must be verified
             GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
lastmjs, line 54: What's the difference between the GitHub Token and the lastmjs github token?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+            GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }} # All commits must be verified
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        outputs:
+            release-version: ${{ steps.get-version.outputs.release-version }}
+            test-infos: ${{ steps.get-test-infos.outputs.test-infos }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.pull_request.head.ref || github.ref }}
+                  token: ${{ secrets.LASTMJS_GITHUB_TOKEN || github.token }}
```
lastmjs, line 28: What's with the or? Not sure we want to not explicitly use the lastmjs token like this

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+        env:
+            GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }} # All commits must be verified
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        outputs:
+            release-version: ${{ steps.get-version.outputs.release-version }}
+            test-infos: ${{ steps.get-test-infos.outputs.test-infos }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.pull_request.head.ref || github.ref }}
```
lastmjs, line 27: Why the ors?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+            - name: Install curl
+              run: sudo apt-get install curl -y
+
+            - id: get-dfx-version
+              uses: ./.github/actions/get_dfx_version
+
+            - name: Install dfx
```
lastmjs, line 52: I feel like we're doing this in a couple places now...could this be made reusable somehow?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+                  echo -n "$GPG_SIGNING_KEY" | base64 --decode | gpg --import
+                  git config --global user.signingkey C8B77BCBE16CD2B94B43F9C8757397B82D4ED7B0
+
+            - name: Update version and build templates
+              run: |
+                  VERSION=${{ steps.get-version.outputs.release-version }}
+                  sed -E -i "s/(\"version\": \")(.*)(\")/\1$VERSION\3/" package.json
+                  sed -E -i "s/(\"version\": \")(.*)(\")/\1$VERSION\3/" dfx_extension/extension.json
+                  npm install
+                  npx azle template
```
lastmjs, line 79: Should we turn on verbose here?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+            LASTMJS_GITHUB_TOKEN:
+                required: true
+
+jobs:
+    prepare-release:
+        name: Prepare Release
+        runs-on: ubuntu-latest
+        env:
+            GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }} # All commits must be verified
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
lastmjs, line 20: Was this token used here?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+            GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        strategy:
+            fail-fast: false
+            matrix:
+                test: ${{ fromJson(needs.prepare-release.outputs.test-infos) }}
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.pull_request.head.ref || github.ref }}
```
lastmjs, line 111: ors?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+            - id: get-dfx-version
+              uses: ./.github/actions/get_dfx_version
+
+            - name: Install dfx
+              run: |
+                  # Install dfx (Note: dfx must be installed before `npx azle` because the azle installation process requires dfx)
+                  src/build/stable/commands/install_global_dependencies/install_dfx.sh ${{ steps.get-dfx-version.outputs.dfx-version }}
+                  echo "$HOME/.local/share/dfx/bin" >> $GITHUB_PATH
+
+            - name: Update dependencies
```
lastmjs, line 130: Can we rename this job to explain what we're updating? To me it was unclear on if this was updating some kind of azle dependencies through installation or something...so it's updating the configurationish files of the examples/tests

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+              run: |
+                  cd ${{ matrix.test.path }}
+                  sed -E -i "s/(\"azle\": \")(.*)(\")/\1${{ needs.prepare-release.outputs.release-version }}\3/" package.json
+                  npm install
+                  rm -rf node_modules
```
lastmjs, line 135: It would be nice to update everything with the dependency including the candid files and benchmarks (benchmarks might be taken care of elsewhere, I just want to discuss doing it here). So if we do a dfx build we could also update any candid files

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ github.event.pull_request.head.ref || github.ref }}
+                  token: ${{ secrets.LASTMJS_GITHUB_TOKEN }}
+                  fetch-depth: 0 # Fetch all history for all branches and tags
+
+            - name: Merge update branches
+              env:
+                  PAT: ${{ secrets.LASTMJS_GITHUB_TOKEN }}
```
lastmjs, line 171: PAT?

.github/workflows/release_parallel.yml
```diff
@@ -0,0 +1,218 @@
...
+                  echo "Current branch: $CURRENT_BRANCH"
+
+                  # Fetch all branches
+                  git fetch --all
+
+                  # List and merge update branches
+                  for branch in $(git branch -r | grep "origin/update-${{ needs.prepare-release.outputs.release-version }}-" | sed 's/origin\///'); do
+                    echo "Merging branch: $branch"
+                    git fetch origin $branch:$branch
+                    git merge --no-ff "$branch" -m "Merge $branch" || {
```
lastmjs, line 190: Are we sure a merge is the best strategy here?

.github/workflows/release_parallel.yml
lastmjs, line 1: There are a lot of "different" GitHub commands in here...have you vetted them each and are you sure they're the simplest way to accomplish what is needed? I feel like a there's a lot of LLM-generated bash/git commands here, which is good but not if we don't know what they're doing and if they're necessary.

lastmjs, line 1: I would love a tour of this new process step by step to help me understand it better.

lastmjs (changes requested):

.github/scripts/publish_github_action.sh
bdemann: I'll do you one better, we can delete the whole file

.github/workflows/test.yml
bdemann: Well... here's the thing, it is shocking to me how many comments AIs will just rip out. I catch most of them but some of the slip through. This is one of those. However, this comment is not really necessary. That is what putting pull_request there mean. And we don't have a corresponding comment on `push:` that says "Runs on pushes to main branch." Anyone that knows the github workflow yml should know that is what it means. And we don't put a comment on a for loop that says "repeats the below code 10 times" because anyone that knows javascript will know that's exactly what a for loop does. So yeah, it's not really necessary

package.json
bdemann: yes I can, I was running into issues where it was trying to do the husky install from this prepare step when you were installing azle into an example or tests. and it was failing because it didn't have husky installed. I was exploring some different options. It seems like the easiest one was to just not do the prepare step if .git/hooks doesn't exists. I'm guessing that the fact that this happened all of a sudden hints at a deeper problem and we should probably talk about it now that you are back from your vacation

.github/workflows/release.yml
bdemann: I don't know, my guess is that the GitHub token is a relic from an old way of doing the workflow. We could test getting rid of it if you'd like

.github/workflows/release_parallel.yml
bdemann: I don't know, it has been that way from time immemorial, dating back to at least 2 years ago when Dan first ran prettier on our code base

.github/workflows/release_parallel.yml
bdemann: same as with the token, it's been like that the whole time. Questioning the old ways of authenticating or checking out was not in the scope of this issue. I thought it best, for time reasons, to only question the old ways of spawning jobs.

.github/workflows/release_parallel.yml
bdemann: it is literally already pulled out to be reusable, the only thing this job does is call a script

.github/workflows/release_parallel.yml
bdemann: That's a good idea

bdemann: done

.github/workflows/release_parallel.yml
bdemann: We have a number of env variables that even if they are not used directly in the yml are used in the tests or else where. It is hard to see just by looking at the yml if this token is used by us, or by an script or something we rely on. We can test a work on getting rid of these relics but it felt outside the scope of this job

.github/workflows/release_parallel.yml
bdemann: relic

.github/workflows/release_parallel.yml
bdemann: Personal Authorization Token

.github/workflows/release_parallel.yml
bdemann: I was until you asked me about it

.github/workflows/release_parallel.yml
bdemann: sure thing

bdemann: I think the tour would help with this

.github/workflows/release_parallel.yml
bdemann: Yep got that here https://github.com/demergent-labs/azle/issues/2173
It's a pretty involved process, this is the first step of several

.github/workflows/release_parallel.yml
bdemann: sure

bdemann: done

package.json
bdemann: Make sure that top level azle is installed before any other npm installs are called

.github/workflows/release.yml
bdemann: let's test getting rid of this

.github/workflows/release_parallel.yml
bdemann: open an issue to explore this for 1.0 milestone

.github/workflows/release_parallel.yml
bdemann: open an issue to explore this for 1.0 milestone

.github/workflows/release_parallel.yml
bdemann: We could thy echoing the path inside the script

.github/workflows/release_parallel.yml
bdemann: Could we explore squashing all of them?

bdemann: could a rebase even without a squash result in half the commits because you won't have each commit followed by a merge commit?

package.json
bdemann: it looks like it was just a network error that interrupted the npm install at the top level

.github/workflows/release_parallel.yml
bdemann: https://github.com/demergent-labs/azle/issues/2200

.github/workflows/release_parallel.yml
bdemann: https://github.com/demergent-labs/azle/issues/2200

.github/workflows/release_parallel.yml
bdemann: https://github.com/demergent-labs/azle/issues/2200

.github/workflows/release.yml
bdemann: gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
Error: Process completed with exit code 4.

.github/workflows/release_parallel.yml
bdemann: done

.github/workflows/release_parallel.yml
bdemann: done

.github/workflows/release_parallel.yml
```diff
@@ -164,46 +140,74 @@ jobs:
...
+                  while read branch; do
+                    echo "  - $branch"
+                  done < <(echo "${{ steps.collect-branches.outputs.branches }}")
+                  echo "End of branch list"
+
+            - name: Fetch branches
+              run: |
+                  echo "Fetching all branches..."
+                  BRANCHES_TO_FETCH=""
+                  while read branch; do
```
lastmjs, line 170: Use for loops and single space separated lists

.github/workflows/release_parallel.yml
```diff
@@ -164,46 +140,74 @@ jobs:
...
-
-                  # List and merge update branches
-                  for branch in $(git branch -r | grep "origin/update-${{ needs.prepare-release.outputs.release-version }}-" | sed 's/origin\///'); do
-                    echo "Merging branch: $branch"
-                    git fetch origin $branch:$branch
-                    git merge --no-ff "$branch" -m "Merge $branch" || {
+                  CURRENT_BRANCH=$(git branch --show-current)
+                  BASE_COMMIT=$(git rev-parse HEAD)
+
+                  while read branch; do
```
lastmjs, line 184: Look into octopus merge or somehow merging all branches at once (git pull origin b1 b2 b3 etc could work?)

lastmjs (changes requested):

.github/scripts/publish_github_action.sh
lastmjs: Please actually delete

.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,43 @@
+name: 'Setup Node.js'
+description: 'Sets up Node.js environment and gets Node version from package.json'
+inputs:
+    working-directory:
```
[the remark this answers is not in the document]
bdemann: done

.github/actions/setup_dfx/action.yml
```diff
@@ -0,0 +1,27 @@
+name: Setup dfx
+description: 'Sets up dfx by detecting version from package.json and installing it. (Note: DFX must be installed before `npm install` because the azle installation process requires dfx)'
+inputs:
+    dfx-version:
```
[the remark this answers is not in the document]
bdemann: done

.github/scripts/publish_github_action.sh
bdemann: done

.github/workflows/release_parallel.yml
```diff
@@ -33,51 +33,32 @@ jobs:
...
                   echo "release-version=$RELEASE_VERSION" >> $GITHUB_OUTPUT
 
-            - id: get-node-version
-              uses: ./.github/actions/get_node_version
-
-            - uses: actions/setup-node@v4
+            - uses: ./.github/actions/setup-node
               with:
-                  node-version: ${{ steps.get-node-version.outputs.node-version }}
                   registry-url: https://registry.npmjs.org
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/release_parallel.yml
```diff
@@ -164,46 +140,74 @@ jobs:
...
-            - name: Merge update branches
+            - uses: ./.github/actions/configure_git
+              with:
+                  gpg_signing_key: ${{ secrets.GPG_SIGNING_KEY }}
+
+            - name: Collect branches
+              id: collect-branches
+              run: |
+                  # Create array of branches
+                  readarray -t BRANCH_ARRAY < <(git branch -r | grep "origin/update-${{ needs.prepare-release.outputs.release-version }}-" | sed 's/origin\///' | xargs -n1)
```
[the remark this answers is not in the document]
bdemann: done

.github/workflows/release_parallel.yml
bdemann: done

.github/workflows/release_parallel.yml
bdemann: I don't think that will be necessary, like I was saying the squash merge is really meant to do just this, you can just keep squashing and then just do one commit at the end without doing anything hacky or weird like a soft reset. It's all very straight forward.

.github/workflows/release_parallel.yml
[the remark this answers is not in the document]
bdemann: Added to https://github.com/demergent-labs/azle/issues/2173

But also it's kind of nice to just throw up a podcast and manually delete branches. It's a very relaxing way to spend the evening. It may be good for our mental health not to automate away such a tranquil experience

lastmjs (approved):

PR #2191, opened 2024-10-18
Run stable on experimental

bdemann:


lastmjs:
We need to reorganize the tests so that we have just a class syntax stable dropdown and a class syntax experimental drop down, or something like that. The idea is that any stable tests running as experimental will be grouped together into their own collection.

lastmjs (approved):

PR #2193, opened 2024-10-18
Rquickjs move crates

lastmjs:


src/build/experimental/commands/compile/get_context.ts
```diff
@@ -32,6 +32,7 @@ export async function getContext(
         consumer,
         managementDid,
         recordBenchmarks:
+            process.env.npm_lifecycle_event === 'pre_tests' ||
```
bdemann, line 35: does it also need `tests`?

src/build/stable/commands/compile/get_context.ts
```diff
@@ -28,6 +28,7 @@ export function getContext(
     const wasmData: WasmData = {
         envVars,
         recordBenchmarks:
+            process.env.npm_lifecycle_event === 'pre_tests' ||
```
bdemann, line 31: what about tests?

bdemann (changes requested):
See other pr for more comments

bdemann (dismissed):

PR #2249, opened 2024-10-29
Remove unnecessary experimental

lastmjs:


src/lib/experimental/error.ts
```diff
@@ -1,10 +1,11 @@
-// import { trap } from './ic_apis/trap'; // TODO why does this break bitcoin_psbt examples? https://github.com/demergent-labs/azle/issues/2003
 import { trap } from './ic/trap';
 
 export function handleUncaughtError(rawError: any): never {
-    const error = rawError instanceof Error ? rawError : new Error(rawError);
-
-    const azleError = `Uncaught ${error.name}: ${error.message}${error.stack}`;
-
-    trap(azleError);
+    if (rawError instanceof Error) {
```
bdemann, line 4: Can experimental use the stable trap? Could we have just one handleUncaughtError for both stable and experimental

src/lib/stable/globals.ts
```diff
@@ -108,9 +108,6 @@ if (globalThis._azleInsideCanister === true) {
         }) as any
     };
 
-    globalThis._azleInitCalled = false;
```
bdemann, line 101: why did we get rid of these?

PR #2268, opened 2024-11-06
Fix stable in experimental tests

lastmjs:


src/lib/stable/ic_apis/azle_ic_stable.ts
```diff
@@ -16,12 +16,12 @@ export type AzleIcStable = {
     candidDecode: (candidBytes: Uint8Array) => string;
     candidEncode: (candidString: string) => Uint8Array;
     canisterBalance: () => string;
-    canisterVersion: () => bigint;
+    canisterVersion: () => number;
```
bdemann, line 19: Why did we change these to number?

bdemann (approved):

PR #2275, opened 2024-11-11
update puppeteer to use new html

bdemann:


tests/end_to_end/http_server/fetch_ic/benchmarks.json
lastmjs, line 1: Can we get rid of these benchmarks? They shouldn't be running automatically, unless you put them in here for a reason

tests/end_to_end/http_server/internet_identity/test/tests.ts
lastmjs, line 1: Should we get rid of all test case-specific timeouts here as well?

lastmjs (changes requested):

lastmjs (approved):

PR #2303, opened 2024-12-06
Inspect_bug

bdemann:


bdemann:
I couldn't reproduce the error

Issue #2053, opened 2024-09-01
Azle 0.25.0


lastmjs:
- [ ] New release process, make sure to build the binaries and commit them as part of the release
- [ ] are we going to get rid of the canister path azle/canister and just do generated types? We need to do a PR to the candid repo to get this to work

# Features

# Breaking Changes

PR #2386, opened 2024-12-23
add maintenance workflow

bdemann:
- [x] Breaking changes enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [x] New documentation enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [x] Code tested thoroughly

### DIFF
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -1,4 +1,4 @@
-name: Maintenance
+name: Example/test maintenance (package-lock updating, formatting, linting)
 on:
     workflow_dispatch:
         inputs:
@@ -44,16 +44,14 @@ jobs:
             - uses: ./.github/actions/commit_and_push
               with:
                   branch-name: ${{ needs.create-branch-prefix.outputs.base-branch }}
-                  commit-message: 'Run global prettier and eslint fixes'
+                  commit-message: 'Example/test maintenance: formatting and linting'
                   gpg-signing-key: ${{ secrets.GPG_SIGNING_KEY }}
                   signing-key-id: ${{ inputs.signing-key-id }}
 
             - id: get-test-infos
               uses: ./.github/actions/get_test_infos
               with:
-                  directories: |
-                      ./examples
-                      ./tests
+                  directories: ./examples
 
     update-test-dependencies:
         needs:
@@ -93,21 +91,20 @@ jobs:
             - uses: ./.github/actions/commit_and_push
               with:
                   branch-name: ${{ steps.create-branch-name.outputs.branch-name }}
-                  commit-message: 'Update dependencies for ${{ matrix.test.displayPath }}'
+                  commit-message: 'Example/test maintenance: update dependencies'
                   gpg-signing-key: ${{ secrets.GPG_SIGNING_KEY }}
                   create-branch: 'true'
                   signing-key-id: ${{ inputs.signing-key-id }}
 
     squash-branches:
         needs:
-            - prepare-maintenance
             - update-test-dependencies
             - create-branch-prefix
         uses: ./.github/workflows/squash_branches.yml
         with:
             base-branch: ${{ needs.create-branch-prefix.outputs.base-branch }}
             branch-prefix: ${{ needs.create-branch-prefix.outputs.branch-prefix }}
-            commit-message: 'Maintenance: Update dependencies and fix formatting'
+            commit-message: 'Example/test maintenance: update dependencies'
             signing-key-id: ${{ inputs.signing-key-id }}
         secrets:
             GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
@@ -116,7 +113,6 @@ jobs:
 
     create-pr:
         needs:
-            - prepare-maintenance
             - squash-branches
             - create-branch-prefix
         runs-on: ubuntu-latest
@@ -132,7 +128,7 @@ jobs:
                   gh pr create \
                     --base main \
                     --head ${{ needs.create-branch-prefix.outputs.base-branch }} \
-                    --title "Maintenance: Update dependencies and fix formatting" \
+                    --title "Example/test maintenance: update dependencies, formatting, and linting" \
                     --body "Automated PR for maintenance tasks:
                     - Updated package-lock.json files
                     - Fixed formatting with Prettier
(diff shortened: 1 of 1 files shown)


.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
+name: Maintenance
```
lastmjs, line 1: Call it `Examples/tests maintenance (npm install, prettier, eslint, etc)`

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+            - uses: actions/checkout@v4
+              with:
+                  ref: ${{ needs.create-branch-prefix.outputs.base-branch }}
+                  token: ${{ secrets.LASTMJS_GITHUB_TOKEN }}
+
+            - uses: ./.github/actions/setup_node
+
+            - run: npm install
+
+            # Run global prettier/lint fixes
```
lastmjs, line 37: Use our new fancy scripts to do the pretty and lint stuff

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+            - name: Run Prettier
+              run: npx prettier --write .
+
+            - name: Run ESLint
+              run: npx eslint . --fix
+
+            - uses: ./.github/actions/commit_and_push
+              with:
+                  branch-name: ${{ needs.create-branch-prefix.outputs.base-branch }}
```
lastmjs, line 46: Should this output be specified in the create-branch-prefix step?

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+                  commit-message: 'Run global prettier and eslint fixes'
+                  gpg-signing-key: ${{ secrets.GPG_SIGNING_KEY }}
+                  signing-key-id: ${{ inputs.signing-key-id }}
+
+            - id: get-test-infos
+              uses: ./.github/actions/get_test_infos
+              with:
+                  directories: |
```
lastmjs, line 54: These are wrong

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+            - name: Run Prettier
+              run: npx prettier --write .
+
+            - name: Run ESLint
+              run: npx eslint . --fix
+
+            - uses: ./.github/actions/commit_and_push
+              with:
+                  branch-name: ${{ needs.create-branch-prefix.outputs.base-branch }}
+                  commit-message: 'Run global prettier and eslint fixes'
```
lastmjs, line 47: Change the commit message to the new name of the workflow

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+              uses: ./.github/actions/get_test_infos
+              with:
+                  directories: |
+                      ./examples
+                      ./tests
+
+    update-test-dependencies:
+        needs:
+            - prepare-maintenance
+            - create-branch-prefix
```
lastmjs, line 61: Remove dependency on create-branch-prefix, do for all

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+on:
+    workflow_dispatch:
+        inputs:
+            signing-key-id:
+                required: true
+                type: string
+                default: C8B77BCBE16CD2B94B43F9C8757397B82D4ED7B0
+
+jobs:
+    create-branch-prefix:
```
lastmjs, line 11: Shouldn't there be outputs here?

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+              working-directory: ${{ matrix.test.path }}
+              run: |
+                  rm -f package-lock.json
+                  npm install
+
+            - name: Create branch name
+              id: create-branch-name
+              uses: ./.github/actions/create_branch_name
+              with:
+                  prefix: ${{ needs.create-branch-prefix.outputs.branch-prefix }}
```
lastmjs, line 90: Is this output up there?

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,139 @@
...
+                  ref: ${{ github.ref }}
+
+            - name: Create Pull Request
+              env:
+                  GH_TOKEN: ${{ secrets.LASTMJS_GITHUB_TOKEN }}
+              run: |
+                  gh pr create \
+                    --base main \
+                    --head ${{ needs.create-branch-prefix.outputs.base-branch }} \
+                    --title "Maintenance: Update dependencies and fix formatting" \
```
lastmjs, line 135: Change all title, names, commit messages to the same thingish

lastmjs (changes requested):

.github/workflows/maintenance.yml
bdemann: done

.github/workflows/maintenance.yml
bdemann: Apparently when you do a reusable workflow like this you CANNOT specify the outputs except in the workflow yml.

.github/workflows/maintenance.yml
bdemann: I changed my mind, these are great and maybe even preferable

.github/workflows/maintenance.yml
bdemann: apparently not

.github/workflows/maintenance.yml
bdemann: done

.github/workflows/maintenance.yml
bdemann: It shouldn't be

.github/workflows/maintenance.yml
bdemann: Done

.github/workflows/maintenance.yml
bdemann: Basically there are two reasons to specify a dependency. One is to establish the run order of the jobs, the other is to get values from outputs. If you only care about the first then you only need the one dependency that run directly before it, but to get the second you need to specify all the dependencies you want the outputs from. I simplified all the ones I could, and left the rest

.github/workflows/maintenance.yml
bdemann: done \

lastmjs (approved):

PR #2398, opened 2024-12-30
update rust

bdemann:
- [x] Breaking changes enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053) (breaking changes need to be added to the issue, at least .raw)
- [x] New documentation enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053) (at least the .raw stuff needs to be added to documentation, should be added to the issue)
- [x] Code tested thoroughly (dfx will automatically have tests run against it, and node to some extent...but how are we testing wasi2ic and rust? We would have to simply run those locally and create new templates right? I assume that's what you did?)
- [x] All tasks in linked issues have been completed or made into separate issues


lastmjs (approved):

bdemann:
- .raw has been added to documentation and breaking change. Note: that's only a breaking change if we can get the dfx version updated. 
- You're assumptions about the building templates to test rust and wasi2ic are correct. As I was double checking I realized that I had not built the template after updating wasi2ic. I have updated that. Strangely the template didn't change after that update...

bdemann:
you may be interested in reading this issue https://github.com/demergent-labs/azle/issues/2403

lastmjs (dismissed):

lastmjs (approved):

PR #2415, opened 2025-01-01
Canister methods only from export

lastmjs:
- [x] Breaking changes enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [x] New documentation enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [x] Code tested thoroughly
- [x] Related issues have been linked and all tasks have been completed or made into separate issues


src/lib/stable/canister_methods/heartbeat.ts
```diff
@@ -2,7 +2,7 @@ import { decoratorArgumentsHandler, MethodType } from '.';
 
 export function heartbeat<This, Args extends any[], Return>(
     originalMethod: MethodType<This, Args, Return>,
-    context: ClassMethodDecoratorContext
+    context: ClassMethodDecoratorContext<This, MethodType<This, Args, Return>>
```
bdemann, line 5: I'm curious abut these types but that's not important

src/lib/stable/canister_methods/index.ts
```diff
@@ -62,104 +80,151 @@ export function decoratorArgumentsHandler<This, Args extends any[], Return>(
...
+        }
+
+        this._azleCallbacks[indexString] = async (
+            args?: Uint8Array
+        ): Promise<void> => {
+            try {
+                await executeAndReplyWithCandidSerde(
+                    canisterMethodMode,
+                    args ?? new Uint8Array(),
+                    originalMethod.bind(this), // TODO manually test context and ensure we have automatic tests for this
```
bdemann, line 216: Does this need to be part of this pr? Or is there an issue for it somewhere?

src/lib/stable/canister_methods/index.ts
```diff
@@ -62,104 +80,151 @@ export function decoratorArgumentsHandler<This, Args extends any[], Return>(
...
+                    paramIdlTypes ?? [],
+                    returnIdlType,
+                    options?.manual ?? false
+                );
+            } catch (error: any) {
+                handleUncaughtError(error);
+            }
+        };
+
+        globalThis._azleExportedCanisterClassInstance = this;
```
bdemann, line 226: If you have multiple classes won't this result in the last class executed being the class that represents the canister?

src/lib/stable/canister_methods/index.ts
```diff
@@ -62,104 +80,151 @@ export function decoratorArgumentsHandler<This, Args extends any[], Return>(
...
-        });
-
-        globalThis._azleCanisterMethodIdlTypes[name] = IDL.Func(
-            paramIdlTypes ?? [],
-            returnIdlType === undefined ? [] : [returnIdlType],
-            ['query']
-        );
-    }
+    context.addInitializer(function () {
+        if (this._azleCanisterMethodsIndex === undefined) {
```
bdemann, line 112: Why do you need to check if they are undefined instead of just initializing them?

src/lib/stable/canister_methods/index.ts
```diff
@@ -62,104 +80,151 @@ export function decoratorArgumentsHandler<This, Args extends any[], Return>(
...
-            composite: options?.composite ?? false
-        });
-
-        globalThis._azleCanisterMethodIdlTypes[name] = IDL.Func(
-            paramIdlTypes ?? [],
-            returnIdlType === undefined ? [] : [returnIdlType],
-            ['query']
-        );
-    }
+    context.addInitializer(function () {
```
bdemann, line 111: If I understand correctly, for this file at least this is the biggest change is that everything happens inside this initializer right? Then instead of modifying the global stuff we modify the class stuff and then just add this stuff all in one go to the global stuff? Can you go over with me one more time what exactly the add Initializer function does?

src/lib/stable/canister_methods/index.ts
bdemann: Or will the default exported class always be the last to finish up since this happens at the end and any other class that might be initialized in pursuit of initializing this class will all be called and done before we get to this line and the default exported class will always ultimately be the last to write to this variable?

src/build/stable/commands/compile/javascript.ts
```diff
@@ -25,24 +25,38 @@ function getPrelude(main: string): string {
...
             return JSON.stringify({
                 candid: toDidString(candid),
-                methodMeta: globalThis._azleMethodMeta
+                methodMeta: exportedCanisterClassInstance._azleMethodMeta
             });
         };
+        
+        function getExportedCanisterClassInstance() {     
+            try {
+                new Canister.default();
```
bdemann, line 46: And the reason we need to initialize the class every time even in the node environment is because even if the decorators always do their thing it's not until the class is initialized that the values actually get set in the globals. Is that right?

src/build/experimental/commands/compile/wasm_binary/rust/experimental_canister_template/src/execute_method_js.rs
```diff
@@ -13,7 +13,16 @@ pub extern "C" fn execute_method_js(function_index: i32, pass_arg_data: i32) {
 
         runtime.run_with_context(|context| {
             let global = context.get_global();
-            let callbacks = global.get("_azleCallbacks");
+
+            let exported_canister_class_instance = global.get("_azleExportedCanisterClassInstance");
```
bdemann, line 17: Is this for if you are using the class syntax in experimental? It's got to be right?

src/build/stable/commands/compile/wasm_binary/rust/stable_canister_template/src/execute_method_js.rs
```diff
@@ -29,14 +29,24 @@ fn execute_method_js_with_result(
     pass_arg_data: bool,
 ) -> Result<(), Box<dyn Error>> {
     quickjs_with_ctx(|ctx| {
-        let callbacks: Object = ctx
+        let exported_canister_class_instance: Object = ctx
```
bdemann, line 32: And this is exactly the same except we are getting the exported canister class instance and *then* getting the callbacks off of that right?

src/lib/stable/canister_methods/index.ts
```diff
@@ -1,55 +1,73 @@
...
+import { MethodMeta } from '../../../build/stable/utils/types';
 import { handleUncaughtError } from '../error';
 import {
     CanisterMethodMode,
     executeAndReplyWithCandidSerde
 } from '../execute_with_candid_serde';
 
+export interface ExportedCanisterClass {
+    _azleCallbacks?: {
+        [key: string]: MethodType<ExportedCanisterClass, any[], any>;
```
bdemann, line 12: Should these be unknown instead of any?

lastmjs: Addressed

src/lib/stable/canister_methods/index.ts
lastmjs: Addressed, please scrutinize deeply my solution, considering global state issues as multiple instances might be created during class constructor or variable instantiation

examples/experimental/test/end_to_end/candid_rpc/timers/src/timers.ts
```diff
@@ -86,7 +86,7 @@ export default Canister({
             repeatCrossCanister: repeatCrossCanisterId
         };
     }),
-    statusReport: query([], StatusReport, () => {
+    getStatusReport: query([], StatusReport, () => {
```
bdemann, line 89: I think this is the right change, I'm just curious why it happened in this pr

examples/stable/test/end_to_end/candid_rpc/complex_types/src/index.ts
```diff
@@ -3,23 +3,31 @@ import { IDL, query, update } from 'azle';
 import { Post, Reaction, ReactionType, Thread, User } from './candid_types';
 import * as posts from './posts';
 import * as reactions from './reactions';
+import { State } from './state';
 import * as threads from './threads';
 import * as users from './users';
 
-export default class {
+export default class Canister {
```
bdemann, line 10: Was this changed to be more classy or did it break with one of the changes?

bdemann: I see that there was a TODO for it

src/lib/stable/canister_methods/index.ts
```diff
@@ -6,50 +6,54 @@ import {
...
-    originalMethod: MethodType<This, Args, Return>,
-    context: ClassMethodDecoratorContext<This, MethodType<This, Args, Return>>
-) => MethodType<This, Args, Return>;
+export type DecoratorFunction<This, Args extends unknown[], Return> = (
+    originalMethod: OriginalMethod<This, Args, Return>,
+    context: Context<This, Args, Return>
+) => void;
 
-export function decoratorArgumentsHandler<This, Args extends any[], Return>(
+export type OriginalMethod<This, Args extends unknown[], Return> = MethodType<
```
bdemann, line 33: If I'm reading this right this is just a rename to help with being declarative right?

src/lib/stable/canister_methods/index.ts
```diff
@@ -6,50 +6,54 @@ import {
...
 ) => Return;
 
-export type DecoratorFunction<This, Args extends any[], Return> = (
-    originalMethod: MethodType<This, Args, Return>,
-    context: ClassMethodDecoratorContext<This, MethodType<This, Args, Return>>
-) => MethodType<This, Args, Return>;
+export type DecoratorFunction<This, Args extends unknown[], Return> = (
+    originalMethod: OriginalMethod<This, Args, Return>,
+    context: Context<This, Args, Return>
+) => void;
```
bdemann, line 31: I presume that after investigating why inspect_message and heartbeat returned void while the other methods returned MethodType you discovered that all of them could return void?

bdemann: I am reviewing as if that assumption is true

src/lib/stable/canister_methods/index.ts
```diff
@@ -6,50 +6,54 @@ import {
     CanisterMethodMode,
     executeAndReplyWithCandidSerde
 } from '../execute_with_candid_serde';
+import { QueryOptions } from './query';
+import { UpdateOptions } from './update';
 
 export interface ExportedCanisterClass {
     _azleCallbacks?: {
-        [key: string]: MethodType<ExportedCanisterClass, any[], any>;
+        [key: string]: (args?: Uint8Array) => Promise<void>;
```
bdemann, line 14: What's going on here?

src/lib/stable/canister_methods/index.ts
```diff
@@ -6,50 +6,54 @@ import {
...
+) => void;
 
-export function decoratorArgumentsHandler<This, Args extends any[], Return>(
+export type OriginalMethod<This, Args extends unknown[], Return> = MethodType<
+    This,
+    Args,
+    Return
+>;
+
+export type Context<
```
bdemann, line 39: Is this also just a declarative wrapper?

bdemann: Do we ever have any Contexts that should be ExportedCanisterClass instead of This?

bdemann (dismissed):

src/lib/stable/canister_methods/index.ts
```diff
@@ -42,6 +42,15 @@ export type Context<
     Return
 > = ClassMethodDecoratorContext<This, MethodType<This, Args, Return>>;
 
+/**
+ * Handles decorator arguments for canister methods, supporting both overloaded and non-overloaded decorators.
+ *
```
bdemann, line 47: How do we feel about using the `@internal` tag

src/lib/stable/canister_methods/inspect_message.ts
```diff
@@ -1,6 +1,24 @@
 import { Context, decoratorArgumentsHandler, OriginalMethod } from '.';
 
-// TODO explain here in a jsdoc that the dev can get the raw args using argDataRaw
+/**
+ * Decorator to mark a method as the inspectMessage method entry point.
+ *
+ * The inspectMessage method entry point will be called just before a call to an update method entry point.
+ *
+ * Arguments to the update method entry point can be accessed using `argDataRaw`.
```
bdemann, line 8: if we are putting a reference to `argDataRaw` do we also want to put a reference to `methodName` and `acceptMessage` conversely if we aren't do the others should we also not do `argDataRaw` or is this too nit picking for release candidate

bdemann (dismissed):

bdemann (approved):

PR #2653, opened 2025-02-05
restore module and update to ES2022

bdemann:
## Contributor

- [ ] Code has been declaratized
- [ ] All new functions have JSDoc/Rustdoc comments
- [ ] Error handling beautiful (no unwraps or expects etc)
- [ ] Code tested thoroughly
- [ ] PR title:
    - [ ] Sentence cased
    - [ ] Described well for release notes
    - [ ] Indicates breaking changes with suffix "(breaking changes)"
- [ ] Related issues have been linked and all tasks have been completed or made into separate issues
- [ ] New documentation enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [ ] All breaking changes
    - [ ] Described below in the "Breaking Changes" section
    - [ ] Migration path described
- [ ] Review is requested when ready

## Reviewer

- [ ] Code has been declaratized
- [ ] All new functions have JSDoc/Rustdoc comments
- [ ] Error handling beautiful (no unwraps or expects etc)
- [ ] Code tested thoroughly
- [ ] PR title:
    - [ ] Sentence cased
    - [ ] Described well for release notes
    - [ ] Indicates breaking changes with suffix "(breaking changes)"
- [ ] Related issues have been linked and all tasks have been completed or made into separate issues
- [ ] New documentation enumerated in [the release issue](https://github.com/demergent-labs/azle/issues/2053)
- [ ] All breaking changes
    - [ ] Described below in the "Breaking Changes" section
    - [ ] Migration path described

## Breaking Changes

- None


http_client/fetch_ic/call.ts
```diff
@@ -126,7 +126,7 @@ async function prepareRequestBody(
         init.body instanceof Float32Array ||
         init.body instanceof Float64Array
     ) {
-        return new Uint8Array(init.body);
+        return new Uint8Array(init.body as ArrayBuffer);
```
lastmjs, line 129: Let's talk about this, if you just get rid of BigUint64Array and BigInt64Array in the union the error gets fixed. We should consider just doing that, and maybe this is some bug we should open an issue for somewhere.

.github/workflows/test.yml
```diff
@@ -151,6 +151,23 @@ jobs:
...
+            - uses: ./.github/actions/retry_command
+              with:
+                  command: 'npm install'
+
+            - run: npm run typecheck
+            - run: npm run lint
+
     check-test-success:
         name: Check Azle tests succeeded
         needs: run-tests
```
lastmjs, line 173: But we want top-level-checks to be hooked up here don't we? We need to make sure there's only pass or fail if run-tests and top-level-checks pass or fail...right?

src/lib/experimental/fetch/http.ts
```diff
@@ -202,7 +202,7 @@ async function prepareRequestBody(
 
     if (typeof init.body === 'string') {
         const textEncoder = new TextEncoder();
-        return [textEncoder.encode(init.body)];
+        return [textEncoder.encode(init.body) as Uint8Array<ArrayBuffer>];
```
lastmjs, line 205: Let's discuss these casts

src/lib/experimental/fetch/icp.ts
```diff
@@ -61,7 +61,7 @@ export async function fetchIcp(
         }
     );
 
-    const decodedResult = IDL.decode(funcIdl.retTypes, result);
+    const decodedResult = IDL.decode(funcIdl.retTypes, result.buffer);
```
lastmjs, line 64: I really want to understand this. Let's discuss why this worked before, because it doesn't make sense to me fully why it would not continue to work.

src/lib/stable/ic_apis/is_controller.ts
```diff
@@ -21,8 +21,10 @@ export function isController(principal: Principal): boolean {
     }
 
     if (globalThis._azleIcExperimental !== undefined) {
+        const principalBytes =
+            principal.toUint8Array() as Uint8Array<ArrayBuffer>;
```
lastmjs, line 25: Whoa let's discuss this, I think there's a better way than these casts, I hope

src/lib/stable/ic_apis/msg_arg_data.ts
```diff
@@ -15,12 +15,12 @@
...
  *   - after a successful inter-canister await from a composite query
  */
-export function msgArgData(): Uint8Array {
+export function msgArgData(): Uint8Array<ArrayBuffer> {
     if (
         globalThis._azleIcStable === undefined &&
         globalThis._azleIcExperimental === undefined
     ) {
-        return new Uint8Array();
+        return new Uint8Array(new ArrayBuffer(0));
```
lastmjs, line 23: Seriously, this has to be done now???????

lastmjs: We can't at least leave off the 0?

tsconfig.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "strict": true,
         "target": "ES2024",
+        "module": "ES2022",
```
lastmjs, line 5: Module and moduleResolution are being taken care of in another PR?

lastmjs: Why ES022 now?

tsconfig.typecheck.json
```diff
@@ -0,0 +1,8 @@
+{
```
lastmjs, line 1: Please help me to understand this

package.json
```diff
@@ -3,7 +3,7 @@
     "version": "0.25.0",
     "description": "TypeScript and JavaScript CDK for the Internet Computer",
     "scripts": {
-        "typecheck": "tsc --noEmit",
+        "typecheck": "tsc --project tsconfig.typecheck.json",
```
lastmjs, line 6: Please help me to understand this

test/index.ts
```diff
@@ -2,9 +2,14 @@ import * as dns from 'node:dns';
 dns.setDefaultResultOrder('ipv4first');
 
 import { describe, expect, test } from '@jest/globals';
+import { execSync } from 'child_process';
```
lastmjs, line 5: Why are we using execSync?

test/index.ts
```diff
@@ -36,12 +41,11 @@ export function runTests(
 
     if (shouldRunTypeChecks === true) {
         describe(`type checks`, () => {
-            it('checks types', () => {
+            it('checks types', async () => {
+                const typeCheckCommand = `npm exec --offline tsc -- --noEmit --skipLibCheck`;
```
lastmjs, line 45: Shouldn't we try to explicitly pass in the tsconfig.json of the project? Just like we do at the top level?

lastmjs (changes requested):

.github/workflows/test.yml
bdemann: Correct, done

http_client/fetch_ic/call.ts
bdemann: Maybe we should still file that issues somewhere?

package.json
bdemann: I don't want to type check the examples at this point. I was figuring that we didn't want to exclude the examples in the main tsconfig.json because it already had an exclude property that was only excluding the new examples (which actually needs to be updated? or gotten rid of, because it's in a new location. (just let me know which one that is and I will do it, otherwise I will add it to the issue about making the ideal tsconfig (another option is that we actually do just exclude the examples in the main ts config? I'm not sure how that would play off extending the main one though...))). Unfortunately you cannot, to my knowledge, pass an ignore list as a parameter, only in the tsconfig.json. You can make another tsconfig and point it tsc to it as a separate project. This allows us to have one tsconfig for building and one for typechecking.

src/lib/experimental/fetch/http.ts
bdemann: Yes let's

src/lib/stable/ic_apis/is_controller.ts
bdemann: Okay let's discuss it

src/lib/stable/ic_apis/msg_arg_data.ts
bdemann: Hmmm... apparently it doesn't have to be done... maybe I did those fixes out of order and it was a type inference problem or something like that from mismatched return types?

test/index.ts
bdemann: I don't think so, I felt like passing in at the top level was a work around we had to do because of the extra files we wanted to ignore. I think this is a much better method under normal circumstances

tsconfig.json
bdemann: Because ES2022 is what we agreed on for module... I thought.

Yes module and module resolution are in a separate [issue](https://github.com/demergent-labs/azle/issues/2661). Though es2022 is the one I am planning on using for that pr.

tsconfig.typecheck.json
bdemann: As mentioned in the package.json comment this is so that we can exclude all of the examples from the type checks. As long as we have to make a separate config for that I figured it would be good to add the other type checking settings like skipLibCheck and noEmit. We extend from the base tsconfig to make sure we use the same settings

test/index.ts
bdemann: I tried using execSyncPretty but it seems like I ultimately needed {encoding: 'utf-8'}. we can add that to execSyncPretty if you want, but I wasn't sure if that would be worth it. If it is we can decide if we just want it to always be that or if we want to add another parameter or convert stdio to an options object like in execSync

bdemann: I have my reasons, but let's not worry about that, I'll just switch it to execSyncPretty

package.json
bdemann: Alternatively if we can safely ignore all of the examples in the main tsconfig then we could use just one tsconfig for the whole project and add the noEmit and skip lib check to this command

examples/stable/test/end_to_end/candid_rpc/complex_init/src/complex_init/index.ts
```diff
@@ -13,7 +13,7 @@ export default class {
     init(): void {
         const tuple = IDL.decode(
             [IDL.Tuple(IDL.Text, User)],
-            msgArgData()
+            new Uint8Array(msgArgData()).buffer
```
lastmjs, line 16: Show me the runtime example, this seems crazy

src/lib/experimental/fetch/icp.ts
```diff
@@ -61,7 +61,8 @@ export async function fetchIcp(
         }
     );
 
-    const decodedResult = IDL.decode(funcIdl.retTypes, result);
+    const resultBytes: Uint8Array<ArrayBuffer> = new Uint8Array(result);
+    const decodedResult = IDL.decode(funcIdl.retTypes, resultBytes.buffer);
```
lastmjs, line 65: Make this like the others, : ArrayBuffer and use .buffer on the right hand side

src/lib/experimental/fetch/response.ts
```diff
@@ -37,7 +37,12 @@ export class AzleFetchResponse {
             return new ArrayBuffer(0);
         }
 
-        return (this.body as unknown as Uint8Array | Buffer).buffer;
+        const body = this.body as unknown as Uint8Array | Buffer;
```
lastmjs, line 40: Try this instead:

```typescript
        return new Uint8Array(this.body as unknown as Uint8Array | Buffer)
            .buffer;
```

lastmjs (changes requested):

lastmjs (approved):

src/lib/experimental/fetch/response.ts
bdemann: Done

src/lib/experimental/fetch/icp.ts
bdemann: Done

examples/stable/test/end_to_end/candid_rpc/complex_init/src/complex_init/index.ts
bdemann: Done

